PR #36, opened 2021-04-05
Feature/WCMSRD-11558 generate image pdf

rshelnutt:
Add ability to download generated maps as either an image or a PDF.


danielimmke (approved):
Looks good!

PR #82, opened 2021-08-02
WCMSRD-12236 fixing issue where changing chart type removes the "hori…

cooms13:
…zontal" or "vertical" state attributes.

Added the dataDescription to the list of default fields when you changes from one chart type to another.


danielimmke (approved):
Good catch

PR #197, opened 2022-04-15
Feature/wcmsrd 13525b

Atash3000:


packages/data-bite/src/components/CircleCallout.js
```diff
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, memo, useContext } from 'react'
+import React from 'react'
```
adamdoe, line 1: @Atash3000 - thanks for the cleanup here!

packages/data-bite/src/CdcDataBite.tsx
```diff
@@ -229,22 +310,20 @@ const CdcDataBite = (
         dataBite = Math.max(...numericalData);
         break;
       case DATA_FUNCTION_MIN:
-        dataBite = Math.min(...numericalData);
+        dataBite =Math.min(...numericalData);
         break;
       case DATA_FUNCTION_MODE:
-        dataBite = getMode(numericalData).join(', ');
+        dataBite = getMode(numericalData).join(''); // Needs to be properly tested !!!
```
adamdoe, line 316: @Atash3000 - this looks like it's working. Thanks for calling it out you can safely remove the comment.

packages/data-bite/src/CdcDataBite.tsx
```diff
@@ -206,11 +281,17 @@ const CdcDataBite = (
...
     let numericalData = []
 
-    //Get the column's data
-    filteredData.forEach(row => {
-      let value = numberFromString(row[dataColumn])
-      if(typeof value === 'number') numericalData.push(value)
-    });
+
+   // Get the column's data
+   // TODO: this  conde line probably will not work properly. Need to be tested
```
adamdoe, line 286: @Atash3000 - Thanks for calling this out. I tested filters on my local and they appear to be working. You can safely remove this comment.

packages/data-bite/src/components/EditorPanel.js
```diff
@@ -32,7 +32,7 @@ const TextField = memo(({label, section = null, subsection = null, fieldName, up
     if('number' !== type || min === null){
       setValue(e.target.value);
     } else {
-      if(!e.target.value || ( parseFloat(min) <= parseFloat(e.target.value ) & parseFloat(max) >= parseFloat(e.target.value))) {
+      if(!e.target.value || ( parseFloat(min) <= parseFloat(e.target.value ) && parseFloat(max) >= parseFloat(e.target.value))) {
```
adamdoe, line 35: @Atash3000 - Can you explain the change from bitwise & to logical && here?

Atash3000: I was having issue on when we text on Message field.when we clear text field then we dont see anything.i assumed it could be something there.but i dont why to be honest.

packages/data-bite/src/CdcDataBite.tsx
```diff
@@ -229,22 +310,20 @@ const CdcDataBite = (
...
-        let rangeMax = applyPrecision(numericalData[numericalData.length - 1]);
-
-        if (config.dataFormat.commas) {
-          rangeMin = Number(rangeMin).toLocaleString('en-US');
-          rangeMax = Number(rangeMax).toLocaleString('en-US');
-        }
-
-        dataBite = config.dataFormat.prefix + applyPrecision(rangeMin) + config.dataFormat.suffix + ' - ' + config.dataFormat.prefix + applyPrecision(rangeMax) + config.dataFormat.suffix;
+        const sortedNumArr = [...numericalData].sort((a, b) => a - b);  // create shallow copy and sort
+        let rangeMin =  applyPrecision((sortedNumArr[0]))
```
adamdoe, line 320: @Atash3000 - can you remove the sortedNumArr here and use Math.min and Math.max? If not the way you are using sortedNumArr looks fine.

packages/data-bite/src/components/EditorPanel.js
```diff
@@ -230,10 +230,12 @@ const EditorPanel = memo(() => {
 
   const getColumns = (filter = true) => {
     let columns = {}
-
-    data.map(row => {
-      Object.keys(row).forEach(columnName => columns[columnName] = true)
-    })
+      if(data.length){
```
adamdoe, line 233: @Atash3000 - nice check here, thanks!

packages/data-bite/src/components/EditorPanel.js
```diff
@@ -346,7 +348,7 @@ const EditorPanel = memo(() => {
                     <li className="three-col">
                       <TextField value={config.dataFormat.prefix} section="dataFormat" fieldName="prefix" label="Prefix" updateField={updateField} />
                       <TextField value={config.dataFormat.suffix} section="dataFormat" fieldName="suffix" label="Suffix" updateField={updateField} />
-                      <TextField type="number" value={config.dataFormat.roundToPlace} section="dataFormat" fieldName="roundToPlace" label="Round" updateField={updateField} />
+                      <TextField  type="number" value={config.dataFormat.roundToPlace} section="dataFormat" fieldName="roundToPlace" label="Round" updateField={updateField} min='0' max='99'  />
```
adamdoe, line 351: @Atash3000 - nice addition. We probably don't even need it up to 99 but we'll leave it as is for now until somebody tells us to limit decimals. Good work!

adamdoe (changes requested):
@Atash3000 - Initial code review went great! I love the TS additions (FC/typing), the checks you added for data, and your grit for working through this component without example data. Your hard work is noticed, ty!

packages/data-bite/src/CdcDataBite.tsx
Atash3000: yes i ll remove.originally it was sorting there but sort affects original array and i copyied it. i think Math.max/min ll do better job.other thing is when we have data as string allways we get NaN or simple string concatinations, so there is either we convert all strings to number and then calculate like using reduce method or we ll just assume users ll work with string and number together.

PR #206, opened 2022-04-20
Feature/wcmsrd 12912

Atash3000:
Main bugs fixed. cdcMap.js file was giving errors as well and errors was connected to my ticked so i had to go there as well.

### DIFF
--- a/packages/map/src/components/DataTable.js
+++ b/packages/map/src/components/DataTable.js
@@ -176,9 +176,9 @@ const DataTable = (props) => {
     const newTableColumns = [];
 
     Object.keys(columns).forEach((column) => {
-      if (columns[column].dataTable === true && '' !== columns[column].name) {
+      if (columns[column].dataTable === true && columns[column].name) {
         const newCol = {
-          Header: columns[column].label || columns[column].name,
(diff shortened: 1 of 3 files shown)


adamdoe (approved):
@Atash3000 - nice work here! The only improvement, if you want to make it, is simplifying the geoName lines in `CdcMap.js` - but I'll leave that up to you. It looks like you added more integrity to our files with typechecking which I appreciate. I pulled the feature to my local and tested it and didn't have any issues so I'll merge this in for additional testing. Thank you!

packages/map/src/components/DataTable.js
```diff
@@ -176,9 +176,9 @@ const DataTable = (props) => {
     const newTableColumns = [];
 
     Object.keys(columns).forEach((column) => {
-      if (columns[column].dataTable === true && '' !== columns[column].name) {
+      if (columns[column].dataTable === true && columns[column].name) {
         const newCol = {
-          Header: columns[column].label || columns[column].name,
+          Header: columns[column].label ? columns[column].label : columns[column].name,
```
adamdoe, line 181: @Atash3000 I think this is the same thing, it looks like you swapped the short circuit evaluation for the ternary.

PR #226, opened 2022-05-03
Feature/wcmsrd 13189a

Atash3000:
Added checkboxes for hiding labels and ticks. Ticket number upodated.

### DIFF
--- a/packages/chart/src/CdcChart.tsx
+++ b/packages/chart/src/CdcChart.tsx
@@ -184,9 +184,9 @@ export default function CdcChart(
       });
     }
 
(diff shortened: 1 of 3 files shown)


adamdoe:
@Atash3000 -

This is looking good! I think you're almost there. There are a few errors I'm seeing:

1) When the chart is horizontal, the value axis "hide axis" toggle hides both axes.
2) We need to hide the data label's within barchart when they're below bar or on date/category axis - these will need the same treatment as "hide label"

Atash3000:
> @Atash3000 -
> 
> This is looking good! I think you're almost there. There are a few errors I'm seeing:
> 
> 1. When the chart is horizontal, the value axis "hide axis" toggle hides both axes.
> 2. We need to hide the data label's within barchart when they're below bar or on date/category axis - these will need the same treatment as "hide label"

PR #230, opened 2022-05-09
Feature/wcmsrd 11415 categorical data map

ShawnStringfield:


adamdoe (commented):
Hey @ShawnStringfield - 

It looks like you have a good start here! 

_Reminder_ - please change the branch you're pulling into from `main` to `test` when creating a PR. 

`packages/map/examples/default-test.json` and `index.html` - these files looks good, one way you could improve the `default-test.json` file would be using the dataUrl key in the config instead of having the large list of counties. It's not necessary, but that way other people could use the long list of county data in their sample config files as well, and it cleans up the config some. Even though these are example config files, I'd also recommend renaming it to something helpful for other devs if they want to use it.

Example:  change `"data": [...lots of county data in this array]` to `"dataUrl":  "/examples/your-county-data-here.json"`

I'd also recommend moving `test-data.json` to the `/examples` folder and renaming that file to something more descriptive (ie. reusable county data). Our data folder is more for real data (ie. mapping state abbreviations to state names, geoJSON, etc).

`EditorPanel.js` - This is a bit tricky to see what was updated. It looks like format on save might have formatted ~half the file. I'd recommend reverting your changes here and adding them in again if it's not too much to ask. I noticed some funky symbols being added in (ie. The territories label was changed to: `TE +RRITORIES LABEL`.

I'm not sure if the radio buttons should provide errors, ie. switching from one to the other and back doesn't seem to work. We can dive deeper into that this week if needed. I'd start with the feedback above and if you have questions reach out!

adamdoe:
Hey @ShawnStringfield -

I added styles for radio buttons and pulled test into your branch. The last thing I'm noticing is that when the numeric radio button is selected and you change the legend type the legend type dropdown disappears.

PR #235, opened 2022-05-17
Wcmsrd 13602 missing aria

ShawnStringfield:


packages/chart/src/components/EditorPanel.js
```diff
@@ -554,7 +554,7 @@ const EditorPanel = () => {
       {undefined === config.newViz && config.runtime && config.runtime.editorErrorMessage && <Error /> }
       <button className={displayPanel ? `editor-toggle` : `editor-toggle collapsed`} title={displayPanel ? `Collapse Editor` : `Expand Editor`} onClick={onBackClick}></button>
       <section className={`${displayPanel ? 'editor-panel' : 'hidden editor-panel'}${isDashboard ? ' dashboard': ''}`}>
-        <div className="heading-2">Configure Chart</div>
+        <div role="heading" className="heading-2">Configure Chart</div>
```
adamdoe, line 557: Hey @ShawnStringfield ,

Can you add `aria-level="2"` here? Thanks!

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role#description

packages/dashboard/src/components/Header.js
```diff
@@ -2,7 +2,7 @@ import React, { useContext, memo } from 'react'
 
 const Header = ({preview, setPreview, back, subEditor = null}) => {
   return (
-    <div className="editor-heading">
+    <div role="heading" className="editor-heading">
```
adamdoe, line 5: @ShawnStringfield - Same thing here:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/heading_role#description

adamdoe (changes requested):
Hey @ShawnStringfield -

Can you update t

ShawnStringfield:
Done. thanks @adamdoe

ShawnStringfield:
I have to resolve this conflict then should be good.

PR #243, opened 2022-05-23
Feature/wcmsrd 12689

Atash3000:


adamdoe (approved):
**Testing**
- [x] Verify Feature Requirements In Package
- [x] Verify Feature Requirements in Editor
- [x] Verify Feature Requirements in VM

**Recommendations**
We don't have a prettier file, but I found some of the indentation/spacing off in the files below.
- [ ]  Update formatting/spacing on ValueInput Component (Code Readability)
- [ ] Update formatting/spacing on props where ValueInput is used (Code Readability)
- [ ] Consider destructuring props on Value Input (Code Readability)

**Additional Notes**
This is a nit pic - but it seems the solution is more complex than needed. I like how you created a custom hook in a seperate file and used the useReducer hook. However, I think there are probably simpler solutions to use for the onChange handler.

As always, thanks for the great work Murad!

PR #276, opened 2022-06-17
Note: Please check line 1278

ShawnStringfield:
handleMapTabbing = `#dataTableSection__${state.general.title.replace(/\s/g, "")}`;

This line seems to add nothing of relevance concerning skipping content. Adding a note here just in case I'm wrong or implemented it incorrectly as it stood out to me as something that needed to be checked.


ShawnStringfield:
Files changes not matching

PR #282, opened 2022-06-28
fixed color contrast from failing

ShawnStringfield:


adamdoe (approved):
Hey @ShawnStringfield - 

We do have a prettier/editorconfig PR open for addressing convents/format on save. It looks like single quotes were changed to double quotes in `colorPalettes.js`. No problem for now, but watch the saving on larger files. When large files are changed it can be hard to identify updates. These changes look good though - appreciate it!

PR #290, opened 2022-07-15
Feature/wcmsrd new text elements

ShawnStringfield:


packages/map/src/CdcMap.js
```diff
@@ -1324,10 +1323,16 @@ const CdcMap = ({className, config, navigationHandler: customNavigationHandler,
...
-					<div role='heading' className={'map-title ' + general.headerColor} tabIndex="0">
-						{parse(title)}
+                    <div role='heading' className={'map-title ' + general.headerColor} tabIndex="0">
+                        <sup className="super_title">{general.super_title}</sup>
+						<div>{parse(title)}</div>
 					</div>
-				</header>
+                </header>
+                {general.intro_text &&
+                    <section className="intro_text">
```
adamdoe, line 1332: Consider changing casing

packages/editor/src/scss/main.scss
```diff
@@ -169,3 +169,13 @@
 .cdc-open-viz-module .cove-temp {
   @import '@cdc/core/styles/v2/layout/_tooltip.scss';
 }
+
+  section.intro_text {
+    padding: 15px;
+  }
+
+  section.footnotes {
```
adamdoe, line 177: Since these are likely global styles, I would recommend moving them into core.

adamdoe (changes requested):
Hey Shawn - this looks great! A few things, I try to remove `console.logs` before merging PR's. I also noticed that we typically use camelCase for property names. I'd recommend doing a find and replace to fix those. There were also a few instances where you check for the parents state before displaying the childs. To get around that you can initialize the state or use optional chaining. Finally, we can probably move the styles you added to core since they appear to be used throughout multiple visualizations. 

Great work, and I appreciate the formatting you did on the files!

adamdoe (approved):
Looks great, thanks @ShawnStringfield!

PR #298, opened 2022-07-21
Wcms 14073 cache buster

ShawnStringfield:


packages/editor/src/components/DataImport.js
```diff
@@ -138,13 +138,21 @@ export default function DataImport() {
 
     // Get the raw data as text from the file
     if (null === fileData) {
-      fileSourceType = 'url'
+      const round = 1000 * 60 * 15;
```
adamdoe, line 141: Hey @ShawnStringfield - 

It looks like the variable for URL is working correctly for the prev. 15 minutes. However, I think it's in the wrong spot as I'm not seeing it appended to data viz's on the front end. I think this might actually need to go in a different repo. We can regroup next week on it!

adamdoe:
Hey @ShawnStringfield - closing this PR for now. We can re-open if its still needed in the future.

PR #312, opened 2022-08-02
Chart > apply formatNumber func to ticks to support prefix/suffix

Atash3000:


adamdoe (approved):
Thanks for the PR and testing Murad.

PR #313, opened 2022-08-03
moved bar from clashing with filter

ShawnStringfield:


adamdoe (approved):
Thanks! 

- Thanks for testing on various chart types
- Discussed use of prettier; good that it's being used but difficult to see updates. We will update editorconfig files and prettierrc files in the future.

PR #319, opened 2022-08-15
Chart> fix formatNum fn

Atash3000:


adamdoe (changes requested):
Prefix/Suffix/Commas are not working on axis labels.

PR #358, opened 2022-09-12
repositioned equal number field

ShawnStringfield:


package.json
```diff
@@ -69,6 +69,7 @@
     "publish": "npx lerna publish"
   },
   "dependencies": {
-    "@cdc/core": "^1.1.2"
+    "@cdc/core": "^1.1.2",
+    "@visx/legend": "^2.10.0"
```
adamdoe, line 73: @ShawnStringfield - is this used anywhere in the package? It looks like you added it on this commit but I'm not seeing any code that indicates needing it.

ShawnStringfield: I think this one got added in by mistake... Im not using it anywhere and actually got an error on build when I pulled from test stating this package was missing... I need to remove then retest. thanks

PR #417, opened 2022-10-13
Deleted private example files

mgordon00:


adamdoe:
Looks great @mgordon00 , thank you!

PR #432, opened 2022-10-24
Fixes Tooltip for Mouse Devices

mpallansch:


adamdoe:
@mpallansch - Looks good, tested on extra small devices on desktop. I learned about matchMedia here. Nice solution.

PR #435, opened 2022-10-25
Adds Ability To Limit Dashboard Table Height

mpallansch:


adamdoe:
@mpallansch - looks great! Tested with a data bite and it's working as expected. The only personal/nit pick code changes I could think of were:

- Possible subjective improvement of having the table height / overflow styles moved out into their own object instead of inline. 
- Possible subjective improvement of having the `.table-height-input` class name be more re-usable. If it's concerned with width, maybe there's a better name to use `ie. width-6rem` or a reusable input name `.dashboard-input`. 

Again, very picky at that point. Looks good!

adamdoe (approved):

PR #436, opened 2022-10-25
Dev 785 order reversed on chart

ShawnStringfield:


adamdoe:
Hey @ShawnStringfield -

Can you catch this branch up to test and remove console logs? Thanks!

packages/chart/src/components/LinearChart.tsx
```diff
@@ -58,6 +58,9 @@ export default function LinearChart() {
   let yScale;
   let seriesScale;
 
+  // Shawn: This reversees the chart.-
```
adamdoe, line 61: @ShawnStringfield - I'm also going to cut this comment out. Nothing personal, but the comment doesn't add any useful information here. A developer should be able to see line 62 and know that the keys are being reversed. Comments can be helpful when a developer needs to know WHY something was done a certain way, but here the WHAT/HOW of the code feels very straight forward. Don't let this stop you from writing good comments though! Merging shortly.

PR #456, opened 2022-11-07
moved cacheBustingString to core

mgordon00:


packages/core/helpers/cacheBustingString.js
```diff
@@ -0,0 +1,7 @@
+export default function cacheBustingString(){
+const cacheBustingString = () => {
```
adamdoe, line 2: nit: indent this line

PR #463, opened 2022-11-09
Feature/dev 1227 fixe super title to render html

ShawnStringfield:


adamdoe:
Hey @ShawnStringfield  - 

This appears to be working. A few things though:

Fix: When I use an H1, there's a style issue stemming from `cdc-open-viz-module sup` where top is being used. Instead of using top there, I'd recommend putting a margin bottom on the sup element.
Nit: The sup element is inside of an aria-role heading. Essentially `<h2><sup></sup></h2>` feels structurally weird. Axe Dev Tools doesn't have an issue with it, but I'm wondering if the aria-role should wrap the title below the super title.

packages/map/src/CdcMap.js
```diff
@@ -1533,10 +1533,10 @@ const CdcMap = ({ className, config, navigationHandler: customNavigationHandler,
             className={tooltips.capitalizeLabels ? 'capitalize tooltip' : 'tooltip'}
           />
         )}
-        {state.general.title &&
+        {(state.general.title || general.superTitle) &&
           <header className={general.showTitle === true ? 'visible' : 'hidden'} {...(!general.showTitle || !state.general.title ? { 'aria-hidden': true } : { 'aria-hidden': false })}>
             <div role="heading" className={'map-title ' + general.headerColor} tabIndex="0" aria-level="2">
```
adamdoe, line 1538: @ShawnStringfield - 

- Move aria-level and role to the div on title & verify that looks good.

PR #470, opened 2022-11-17
Feature/dev 836

Atash3000:


packages/map/src/components/EditorPanel.js
```diff
@@ -737,6 +748,7 @@ const EditorPanel = (props) => {
 
 		setRequiredColumns(columnList);
 	}, [state.columns, state.general.type]);
+	console.log(state.legend)
```
adamdoe, line 751: @Atash3000 - remove console logs please.

packages/map/src/components/Sidebar.js
```diff
@@ -114,7 +115,7 @@ const Sidebar = (props) => {
 
   return (
     <ErrorBoundary component="Sidebar">
-      <aside id="legend" className={`${legend.position} ${legend.singleColumn ? 'single-column cdcdataviz-sr-focusable' : 'cdcdataviz-sr-focusable'} ${viewport}`} role="region" aria-label="Legend" tabIndex="0">
+      <aside id="legend" className={`${legend.position} ${className} cdcdataviz-sr-focusable ${viewport}`} role="region" aria-label="Legend" tabIndex="0">
```
adamdoe, line 118: @Atash3000 - good idea condensing this. You could possibly go a little further and put the classes into an array, ie

```
let columnLogic = (legend.position==='side' && legend.singleColumn) ? 'single-column' : (legend.position==='bottom' && legend.singleRow) ? 'single-row' : '';

let classNames = [
'cdcdataviz-sr-focusable',
`${viewport}`,
`${columnLogic}
]
```
and then use `classNames.join(' ')` in the class section

adamdoe:
@Atash3000 -

Overall the code is formatted well and looks good. 

A few things:

**needs attention**: Side legends are not working on maps. Check with the PMs to see if this was ever working.
nitpick: Remove console log
nitpick: Move all classes into an array and use `className.join(' ')` on the div to reduce line length on divs.
nitpick: Use `prepare-commit-msg` found at the root to format commit messages

adamdoe (approved):

PR #549, opened 2023-01-30
Feature/dev 2275 dont plot non numeric values CONT'd

tturnerswdev33:
Finished adding some data scrubbing that will allow line, bar, combo, and pie charts to not break if
users enter these types of data:
null
""
"1,234" - numbers with commas or $ signs; chars stripped and resulting number plotted
"anystring"

- [X] My pull request was branched from and targets the test branch
- [X] I have performed a self-review of my code
- [X] I have manually tested all packages affected (bonus points for automations)
- [ I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] I have checked to ensure there aren't other open pull requests for the same change

- [X] Standalone Component
- [ ] Standalone Full Editor
- [ ] CDC Internal Checks


adamdoe:
@tturnerswdev33 is this ready for review?

packages/chart/src/components/BarChart.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useContext, useState, useEffect } from 'react'
+import React, { useContext, useState, useEffect, useDebugValue } from 'react'
```
adamdoe, line 1: Is `useDebugValue` used anywhere? I'm wondering if we should remove it. Either way, I'm happy to see you used it while working through the ticket. I've never used it, so I was able to learn something new 🙂

packages/chart/src/components/BarChart.tsx
```diff
@@ -9,6 +9,9 @@ import { BarStackHorizontal } from '@visx/shape'
 
 export default function BarChart({ xScale, yScale, seriesScale, xMax, yMax, getXAxisData, getYAxisData, animatedChart, visible }) {
   const { transformedData: data, colorScale, seriesHighlight, config, formatNumber, updateConfig, colorPalettes, formatDate, parseDate } = useContext<any>(Context)
+
+  //debugger;
```
adamdoe, line 13: Nit: try to remove large gaps between variables and commented code when possible. Great use of the debugging tools in this file though. Great to see debugging happening outside of console.log statements. Thank you!

packages/chart/src/components/BarChart.tsx
```diff
@@ -94,6 +97,44 @@ export default function BarChart({ xScale, yScale, seriesScale, xMax, yMax, getX
     return Math.ceil(context.measureText(text).width)
   }
 
+  function cleanData(data) {
```
adamdoe, line 100: nit pick: cleanData makes sense to me right now, but I'm wondering if we could use a different function name here. Or, maybe we should add comments above the function declaration sharing what this function does in simple terms.

packages/chart/src/components/BarChart.tsx
```diff
@@ -94,6 +97,44 @@ export default function BarChart({ xScale, yScale, seriesScale, xMax, yMax, getX
...
+    let cleanedup = []
```
adamdoe, line 101: nit pick: variable name could be improved

packages/chart/src/components/BarChart.tsx
```diff
@@ -135,10 +176,11 @@ export default function BarChart({ xScale, yScale, seriesScale, xMax, yMax, getX
 
   return (
     <ErrorBoundary component='BarChart'>
-      <Group left={config.runtime.yAxis.size}>
+      <Group left={config.runtime.yAxis.size ? parseInt(config.runtime.yAxis.size) : 66}>
```
adamdoe, line 179: Is 66 a magic number here?
https://en.wikipedia.org/wiki/Magic_number_(programming)

This line should probably be reset.

packages/chart/src/components/BarChart.tsx
```diff
@@ -135,10 +176,11 @@ export default function BarChart({ xScale, yScale, seriesScale, xMax, yMax, getX
...
+        {/* left - expects a number not a string */}
         {/* Stacked Vertical */}
         {config.visualizationSubType === 'stacked' && !isHorizontal && (
-          <BarStack data={data} keys={config.runtime.barSeriesKeys || config.runtime.seriesKeys} x={(d: any) => d[config.runtime.xAxis.dataKey]} xScale={xScale} yScale={yScale} color={colorScale}>
+          <BarStack data={cleanData(data)} keys={config.runtime.barSeriesKeys || config.runtime.seriesKeys} x={(d: any) => d[config.runtime.xAxis.dataKey]} xScale={xScale} yScale={yScale} color={colorScale}>
```
adamdoe, line 183: Can we call cleanData before the component?

packages/chart/src/components/BarChart.tsx
```diff
@@ -94,6 +97,44 @@ export default function BarChart({ xScale, yScale, seriesScale, xMax, yMax, getX
...
+          }
+        }
+      })
+      //console.log("cleanedBar=", cleanedBar);
+      cleanedup.push(cleanedBar)
+    })
+    //console.log('## cleanedData =', cleanedup)
+    return cleanedup
+  }
+  const isNumber = value => {
```
adamdoe, line 126: This could be beneficial to other charts so its possible we want it in `core/helpers` or to lift it up to `CdcChart`

packages/chart/src/components/LineChart.tsx
```diff
@@ -36,9 +36,63 @@ export default function LineChart({ xScale, yScale, getXAxisData, getYAxisData,
     return `${formatNumber(value, axis)}`
   }
 
+  const isNumber = value => {
```
adamdoe, line 39: It looks like this is an example where we're seeing repeated code with isNumber.

packages/chart/src/components/LineChart.tsx
```diff
@@ -36,9 +36,63 @@ export default function LineChart({ xScale, yScale, getXAxisData, getYAxisData,
...
+    }
+    return cy
+  }
+
+  // REMOVE bad data points from the data set
+  // Examples: NA, N/A, "1,234", "anystring"
+  // - if you dont call this on data into LineGroup below, for example
+  // then entire data series are removed because of the defined statement
+  // i.e. if a series has any bad data points the entire series wont plot
+  function cleanData(data) {
```
adamdoe, line 64: Repeated fn, if we're re-using this lift this up to CdcChart

packages/chart/src/components/LineChart.tsx
```diff
@@ -91,8 +152,9 @@ export default function LineChart({ xScale, yScale, getXAxisData, getYAxisData,
                       <circle
                         key={`${seriesKey}-${dataIndex}`}
                         r={circleRadii}
-                        cx={xScale(getXAxisData(d))}
-                        cy={seriesAxis === 'Right' ? yScaleRight(getYAxisData(d, seriesKey)) : yScale(getYAxisData(d, seriesKey))}
+                        cx={Number(xScale(getXAxisData(d)))}
+                        //cy={4}
```
adamdoe, line 156: remove

packages/chart/src/components/BarChart.tsx
tturnerswdev33: I can remove it if you want me to.  Sorry forgot that.

packages/chart/src/components/LineChart.tsx
```diff
@@ -121,24 +183,35 @@ export default function LineChart({ xScale, yScale, getXAxisData, getYAxisData,
...
               />
               {config.animate && (
                 <LinePath
                   className='animation'
                   curve={allCurves.curveLinear}
-                  data={data}
+                  data={cleanData(data)}
                   x={d => xScale(getXAxisData(d))}
-                  y={d => (seriesAxis === 'Right' ? yScaleRight(getYAxisData(d, seriesKey)) : yScale(getYAxisData(d, seriesKey)))}
+                  y={d => {
```
adamdoe, line 199: consider using a handler when longer functions are used:

`const handleLinePath = () => {}`
...
`y={handleLinePath}`

Once ternarys get to a certain point in length, in my opinion it feels better to have them moved into there own handler. I'll let you be the judge here.

packages/chart/src/components/PieChart.tsx
```diff
@@ -20,6 +20,36 @@ const enterUpdateTransition = ({ startAngle, endAngle }: PieArcDatum<any>) => ({
   endAngle
 })
 
+// Would put this in helpers but it changes based on the structure of the data
+// - here the pie needs to omit 'name' but line and bar omit "Date"
+function cleanData(data) {
```
adamdoe, line 25: You know the pattern here by now. If these are the same, move them up to CdcChart and pass the fn through.

packages/chart/src/components/BarChart.tsx
tturnerswdev33: This was a red console error occurring not related to the ticket but easy to fix.
THe 66 came from some default setting on yAxis.size that I found somewhere else.
So yes I suppose it is a magic number since one would not know where it came from.
You want it to be a CONSTANT defined somewhere else?

packages/chart/src/components/PieChart.tsx
```diff
@@ -143,6 +173,7 @@ export default function PieChart() {
           newFilteredData.push(d)
         }
       })
+      debugger
```
adamdoe, line 176: remove

packages/chart/src/components/BarChart.tsx
tturnerswdev33: We could call it on data above all the components if you want
data = cleanedData(data)
but in the file the data comes in as a const so you get an error doing that.
Otherwise would have to map it to another variable and then put that variable as input to the components.
Seemed easier just to call it on the inbound data but I will do whatever you want me to do with it.

packages/chart/src/components/SparkLine.js
```diff
@@ -90,10 +90,19 @@ export default function SparkLine({ width: parentWidth, height: parentHeight })
...
 
+  const isNumber = value => {
+    return value !== null && value !== '' && typeof value === 'number'
+  }
+
   useEffect(() => {
     ReactTooltip.rebuild()
   })
 
+  const getYscaleLog = value => {
```
adamdoe, line 101: Great way of handling the debugging. However, we don't want to console log values every time the chart loads. Consider removing or adding a ternary on the log.

packages/chart/src/components/SparkLine.js
```diff
@@ -90,10 +90,19 @@ export default function SparkLine({ width: parentWidth, height: parentHeight })
 
   const handleSparkLineTicks = [xScale.domain()[0], xScale.domain()[xScale.domain().length - 1]]
 
+  const isNumber = value => {
```
adamdoe, line 93: Better naming used in this fn than checkCy in the other file. If they're the same, lift the functions to CdcChart and pass through.

packages/chart/src/components/BarChart.tsx
tturnerswdev33: Yes I thought of that but if you look at the code there is an if statement related to whatever is not data to be checked like "Date" and that changes.  It's been a couple of days so I forget where but it was some other field in another place that isNumber was used so that's why I did not make it a helper.  Another option is have several versions like isNumberExcludeDate and isNumberExcludeSomeotherfield or perhaps change it to pass in the name of the key to exclude: isNumberExcludeKey

packages/chart/src/components/LineChart.tsx
tturnerswdev33: Yep see my other reply

packages/chart/src/components/LineChart.tsx
```diff
@@ -36,9 +36,63 @@ export default function LineChart({ xScale, yScale, getXAxisData, getYAxisData,
...
+    if (typeof value === 'number') {
+      return !Number.isNaN(value)
+    }
+    if (typeof value === 'string') {
+      return value !== null && value !== '' && /\d+\.?\d*/.test(value)
+    }
+    return false // if we get here something is wrong so return false
+  }
+
+  const checkCy = cy => {
```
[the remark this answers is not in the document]
tturnerswdev33: That was mainly for initial debugging bc I thought the cy input was causing it to crash.  It could possibly be removed.  If it's needed I can revisit the name for sure.

packages/chart/src/components/LineChart.tsx
tturnerswdev33: OK sure I can do that.

packages/chart/src/components/LineChart.tsx
tturnerswdev33: I admit I was not sure if that even needed adjusting so was iterating on that... in the end that is when I should have pulled that out into a handler.  No problem I can do that.

packages/chart/src/components/PieChart.tsx
tturnerswdev33: Yep

packages/chart/src/components/PieChart.tsx
tturnerswdev33: Woops

packages/chart/src/components/SparkLine.js
tturnerswdev33: Hmmm didnt realize I changed SparkLine.  I can remove.

packages/chart/src/components/SparkLine.js
tturnerswdev33: yep

adamdoe:
Finally - consider re-setting the example files to the original states.

packages/chart/src/components/BarChart.tsx
adamdoe: Yeah go ahead and remove it, I'm happy it was there for me to learn :)

adamdoe (changes requested):

packages/chart/src/components/BarChart.tsx
tturnerswdev33: I usually add those gaps to that later its supposed to jump out at me to remove... in addition to doing a global searchg check.  ANyway removing

packages/chart/src/components/BarChart.tsx
```diff
@@ -94,6 +97,44 @@ export default function BarChart({ xScale, yScale, seriesScale, xMax, yMax, getX
     return Math.ceil(context.measureText(text).width)
   }
 
+  function cleanData(data) {
+    let cleanedup = []
+    //console.log('## Data to clean=', data)
```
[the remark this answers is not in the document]
tturnerswdev33: Yep I'm changing it to that in case another dev wants to enable and see the difference incoming data vs cleaned

packages/chart/src/components/BarChart.tsx
tturnerswdev33: I thought it was an obvious name... but changing it to cleanedupData

packages/chart/src/components/PieChart.tsx
tturnerswdev33: My exclude filters on global search was excluding .tsx files which is why I was not finding this in a search for debugger statements prior to PR.
Screen Shot 2023-02-02 at 11 39 44 AM

packages/chart/src/components/SparkLine.js
tturnerswdev33: Moved isNumber to CdcChart ... looking at checkCY... cant recall if we need that or not

packages/chart/src/hooks/useReduceData.ts
```diff
@@ -2,8 +2,35 @@ function useReduceData(config, data) {
   // for combo charts check if all  Data Series selected to Bar;
   const isBar = config?.series?.every(element => element?.type === 'Bar')
   // for combo charts check if all Data series selected Line or dashed-md/sm/lg.
-  const isAllLine = config?.series?.every(el => el.type === 'Line' || el.type=== 'dashed-sm'|| el.type=== 'dashed-md' || el.type=== 'dashed-lg');
-
+  const isAllLine = config?.series?.every(el => el.type === 'Line' || el.type === 'dashed-sm' || el.type === 'dashed-md' || el.type === 'dashed-lg')
+  const isNumber = value => {
```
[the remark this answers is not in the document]
tturnerswdev33: THis one is slightly different than the other one. It also cleans out some bad characters before the check rather than having to add clean and isNum in the ternary operators below.

packages/chart/src/hooks/useReduceData.ts
```diff
@@ -2,8 +2,35 @@ function useReduceData(config, data) {
...
+    if (typeof value === 'string') {
+      return value !== null && value !== '' && /\d+\.?\d*/.test(value)
+    }
+    // just in case data has type number we need this
+    if (typeof value === 'number') {
+      return !Number.isNaN(value)
+    }
+    return false // because if it gets here something is wrong
+  }
+  const isNumberLog = value => {
```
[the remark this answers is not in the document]
tturnerswdev33: I'm removing isNumberLog... dont need that to stay in the repo

packages/chart/src/components/BarChart.tsx
tturnerswdev33: Lifted it up to CdcChart

packages/chart/src/components/LineChart.tsx
tturnerswdev33: Lifted it up to CdcChart

packages/chart/src/components/LineChart.tsx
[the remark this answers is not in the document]
tturnerswdev33: Confirmed it's not needed and removed it.

packages/chart/src/components/LineChart.tsx
tturnerswdev33: I had done this early on in the testing of where the problem was... cleaning the data prior to sending it into the component fixes the problem and this extra checking here is redundant so removing it.

packages/chart/src/components/SparkLine.js
tturnerswdev33: Removed this entirely.  Not needed for the repo.

packages/chart/src/components/BarChart.tsx
tturnerswdev33: I'm removing the ternary entirely and just adding parseFloat which is more forgiving than parseInt. Not sure what type of values are going into there.  I tested it and do not get the red error related to setting value "left" on group.

adamdoe:
> My exclude filters on global search was excluding .tsx files which is why I was not finding this in a search for debugger statements prior to PR.
Screen Shot 2023-02-02 at 11 39 44 AM
 

that makes sense, np!

packages/chart/src/components/LineChart.tsx
```diff
@@ -12,7 +12,7 @@ import Context from '../context'
 import useRightAxis from '../hooks/useRightAxis'
 
 export default function LineChart({ xScale, yScale, getXAxisData, getYAxisData, xMax, yMax, seriesStyle = 'Line' }) {
-  const { colorPalettes, transformedData: data, colorScale, seriesHighlight, config, formatNumber, formatDate, parseDate, updateConfig } = useContext<any>(Context)
+  const { colorPalettes, transformedData: data, colorScale, seriesHighlight, config, formatNumber, formatDate, parseDate, isNumber, updateConfig } = useContext<any>(Context)
```
adamdoe, line 15: Perfect, ty!

packages/chart/src/components/BarChart.tsx
```diff
@@ -94,6 +95,36 @@ export default function BarChart({ xScale, yScale, seriesScale, xMax, yMax, getX
     return Math.ceil(context.measureText(text).width)
   }
 
+  // This cleans a data set by:
+  // - removing commas and $ signs from any numbers to try to plot the point
+  // - removing any data points that are NOT composed of of all digits (but allow a decimal point)
+  // Without this the charts "break" and do not render
+  const cleanData = (data, testing = false) => {
```
adamdoe, line 102: Although the functions differ between Pie/Bar/etc charts, I would still try lifting this up to CdcChart. You can take the visualization type as a parameter to run additional logic if needed. Something like:

`const cleanData = (data, testing = false, visType)`

In some visualizations you're logging while in others your not.

packages/chart/src/components/BarChart.tsx
```diff
@@ -135,10 +166,11 @@ export default function BarChart({ xScale, yScale, seriesScale, xMax, yMax, getX
 
   return (
     <ErrorBoundary component='BarChart'>
-      <Group left={config.runtime.yAxis.size}>
+      <Group left={parseFloat(config.runtime.yAxis.size)}>
+        {/* left - expects a number not a string */}
```
adamdoe, line 170: Excellent find, lets remove the comment for readability.

packages/chart/src/components/PieChart.tsx
adamdoe, line 25: See other cleanData fn comment.

packages/chart/src/components/SparkLine.js
```diff
@@ -90,6 +90,10 @@ export default function SparkLine({ width: parentWidth, height: parentHeight })
...
```
adamdoe, line 93: Can you use the isNumber fn you lifted up to CdcChart here?

packages/chart/src/components/SparkLine.js
```diff
@@ -135,6 +139,9 @@ export default function SparkLine({ width: parentWidth, height: parentHeight })
                         style={{ fill: colorScale ? colorScale(config.runtime.seriesLabels ? config.runtime.seriesLabels[seriesKey] : seriesKey) : '#000' }}
                         data-tip={tooltip}
                         data-for={`cdc-open-viz-tooltip-${config.runtime.uniqueId}`}
+                        defined={(item, i) => {
```
adamdoe, line 142: I'm not sure that circles have a defined prop:
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/circle

When its a visx linepath you can use that trick. See ln 114 on LineChart.tsx to see how we're handling text and circles.

packages/chart/src/hooks/useReduceData.ts
```diff
@@ -2,8 +2,22 @@ function useReduceData(config, data) {
   // for combo charts check if all  Data Series selected to Bar;
   const isBar = config?.series?.every(element => element?.type === 'Bar')
   // for combo charts check if all Data series selected Line or dashed-md/sm/lg.
-  const isAllLine = config?.series?.every(el => el.type === 'Line' || el.type=== 'dashed-sm'|| el.type=== 'dashed-md' || el.type=== 'dashed-lg');
-
+  const isAllLine = config?.series?.every(el => el.type === 'Line' || el.type === 'dashed-sm' || el.type === 'dashed-md' || el.type === 'dashed-lg')
+  const isNumber = value => {
```
adamdoe, line 6: Can this be combined with the isNumber lifted to CdcChart? Can we add comments to clarify any regex expressions added. I understand we're returning a true/false value here, but adding clarifying remarks around regex statements would be beneficial so developers don't have to look them up in the future.

packages/chart/src/CdcChart.tsx
```diff
@@ -748,7 +748,17 @@ export default function CdcChart({ configUrl, config: configObj, isEditor = fals
 
   const getXAxisData = (d: any) => (config.runtime.xAxis.type === 'date' ? parseDate(d[config.runtime.originalXAxis.dataKey]).getTime() : d[config.runtime.originalXAxis.dataKey])
   const getYAxisData = (d: any, seriesKey: string) => d[seriesKey]
-
+  const isNumber = value => {
```
adamdoe, line 751: good

packages/chart/src/CdcChart.tsx
```diff
@@ -781,6 +791,7 @@ export default function CdcChart({ configUrl, config: configObj, isEditor = fals
     dynamicLegendItems,
     setDynamicLegendItems,
     filterData,
+    isNumber,
```
adamdoe, line 794: nicely done, follows the conventions of the codebase.

adamdoe (changes requested):
💪 Much better

See the feedback above re: isNumber and cleanData fn's and ping me if you have any questions! Great work!

packages/chart/src/components/BarChart.tsx
tturnerswdev33: Yes I can do that but it makes the inbound data look like:
<BarStack data={cleanData(data,false,"bar")} 
Is that what you want?

packages/chart/src/components/PieChart.tsx
tturnerswdev33: OK done on next push

packages/chart/src/components/SparkLine.js
tturnerswdev33: removing the defined.

packages/chart/src/components/SparkLine.js
tturnerswdev33: Yes

packages/chart/src/hooks/useReduceData.ts
tturnerswdev33: moved to core/helper function

tturnerswdev33: I moved it to core/helpers and imported it here and into CdcChart

tturnerswdev33: and explained the regex in the helper func

PR #551, opened 2023-01-31
Categorical Maps: "Add Category" functionality

mpallansch:

- [ ] Standalone Component
- [ ] Standalone Full Editor
- [ ] CDC Internal Checks

### DIFF
--- a/packages/map/src/CdcMap.js
+++ b/packages/map/src/CdcMap.js
@@ -82,6 +82,15 @@ const hashObj = row => {
   }
 }
 
+const indexOfIgnoreType = (arr, item) => {
+  for(let i = 0; i < arr.length; i++){
+    if(item == arr[i]){
+      return i;
+    }
+  }
+  return -1;
+}
+
 // returns string[]
 const getUniqueValues = (data, columnName) => {
   let result = {}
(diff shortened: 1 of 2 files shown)


packages/map/src/CdcMap.js
```diff
@@ -82,6 +82,15 @@ const hashObj = row => {
   }
 }
 
+const indexOfIgnoreType = (arr, item) => {
```
adamdoe, line 85: This makes sense to me. You might be able to use `Array.prototype.findIndex` here:

`const indexOfIgnoreType = (arr, item) => arr.findIndex(element => element === item);`

A few additional things to keep in mind:
- type checking the array
- use strict equality checks
- variable names (getIndexOf, findIndexOf, etc.), without the ticket in front of me indexOfIgnoreType sounds confusing.

Let me know if you want to make any updates here or if we should just pass it through.

packages/map/src/components/EditorPanel.js
```diff
@@ -1848,6 +1854,57 @@ const EditorPanel = props => {
...
+                              <p>You can provide additional categories to ensure they appear in the legend</p>
+                            </Tooltip.Content>
+                          </Tooltip>
+                        </span>
+                      </label>
+                      {state.legend.additionalCategories && state.legend.additionalCategories.map((val, i) => (
+                        <fieldset className='edit-block' key={val}>
+                          <button
+                            className='remove-column'
+                            onClick={event => {
```
adamdoe, line 1876: Nothing to update here - I just wanted to point out that in a few lines in the codebase, I've been seeing long functions on components. When they start impacting readability I've been moving them into handler functions ie `handleButtonClick`.

adamdoe (changes requested):
I think the code here looks great. When I tested switching the example-city-state map to categorical the editor broke, so maybe run through a few additional scenarios. Im not sure if that was from any of the changes here though. I'll leave it up to you to decide on pulling the code in.

PR #580, opened 2023-02-14
Feature/dev 815 add hide data link checkbox PLUS 3115 checkbox labeling updates

tturnerswdev33:
See tickets 815 and 3115 involving renaming and moving a number of "Show" checkboxes
As well as fixing link for "Go to Table" to show DURING dashboard component config rather than only under "View Dashboard"

Could not get the Go TO Table links on same line as Download CSV link due to being in completely separate components and pulling tableLink into its own component would not fix it bc Download CSV inside the show/hid of Data Table.

Also because this requires so much Editor/Dashboard testing I added a debug functionality that sets the category and data cols on charts and maps by simply using the dataKey for category and then any data col ... this drastically reduces clicking when testing (I have a borderline carpal tunnels situation on my shoulder).  To enable this add: &debug=true after editor=true
(Perhaps we delete this comment from this so it is not generally known???)

This function is only on charts and maps but could be extended to all the components.  Also we could use it to add basic debug logging on components so that we dont have to keep adding and removing console logs every time we work on something... for example log state or config as component is entered.

** This also has CSS fix for the huge help question marks on editor panel... at least for Show Data Table accordian. (there is another ticket to fix them everywhere but if we ca merge this into test then we can take similar approach to all the others)

- [x] Standalone Component
- [x] Standalone Full Editor
- [ ] CDC Internal Checks


tturnerswdev33:
Made some revisions and fixed CSS but I did not see all the issues we saw this morning.
I know that the version you showed me this morning did not even have the checkboxes labeled correctly so not sure what happened there.  I think this is ready to merge but test it again if you want.

packages/dashboard/src/scss/main.scss
```diff
@@ -323,3 +324,7 @@
     margin: 1em 0;
   }
 }
+
+.margin-href {
```
adamdoe, line 328: fix: avoid using !important
fix: use selector inside of the .cdc-open-viz-module namespace, if its within the dashboard nest inside that selector as well.

packages/chart/src/CdcChart.jsx
```diff
@@ -747,6 +731,11 @@ export default function CdcChart({ configUrl, config: configObj, isEditor = fals
   let body = <Loading />
 
   if (!loading) {
+    const tableLink = (
```
adamdoe, line 734: fix: align with additional media links. You might want to use the cove media controls hook here.

adamdoe (changes requested):
Final request, it looks like the link to view the dataset at the component level isn't working.

packages/chart/src/CdcChart.jsx
tturnerswdev33: I did not move that link at all and the ticket did not say to do any alignment changes.  It was already like that.  If you want that changed, how do you want it aligned specifically?

packages/dashboard/src/scss/main.scss
tturnerswdev33: OK yep moved that to _global and gave it a better name.  Technically it could be used for left margin on anything and not just an href... could also make it way longer name like download-link-margin-left if you prefer

packages/core/styles/v2/components/modal.scss
```diff
@@ -67,5 +67,6 @@
 
   .cove-icon {
     flex: 0 0 auto;
+    display: block;
```
adamdoe, line 70: fix: add to the specific selector. This is specific to modals. If you can't find where to put it, place it somewhere inside the editor panel thats specific to the accordion you're working on.

packages/core/styles/v2/components/icon.scss
```diff
@@ -1,5 +1,5 @@
 .cove-icon {
-  display: inline-flex;
+  display: block;
```
adamdoe, line 2: fix: see the comment on modals

packages/chart/src/components/EditorPanel.jsx
```diff
@@ -2250,7 +2251,7 @@ const EditorPanel = () => {
                   {config.table.limitHeight && <TextField value={config.table.height} section='table' fieldName='height' label='Data Table Height' type='number' min='0' max='500' placeholder='Height(px)' updateField={updateField} />}
                   <CheckBox value={config.table.expanded} section='table' fieldName='expanded' label='Expanded by Default' updateField={updateField} />
                   {isDashboard && <CheckBox value={config.table.showDataTableLink} section='table' fieldName='showDataTableLink' label='Show Data Table Name & Link' updateField={updateField} />}
```
adamdoe, line 2255: fix: remove isDashboard check

adamdoe (changes requested):
fix: charts break on the initial render. Please take your time and make sure you go through the PR review template.

packages/core/styles/v2/components/modal.scss
tturnerswdev33: Removed.  This was added as a test to see if it had any effect.  It did not but I admit I forgot to remove it.

packages/core/styles/v2/components/icon.scss
tturnerswdev33: Removed.  This was changed as a quick test to see if it had any effect on question mark styling.  It did not but I admit I forgot to remove it.

tturnerswdev33:
> fix: charts break on the initial render. Please take your time and make sure you go through the PR review template.

Was mainly working in the full Editor and had to run out to take MIL to chiro appointment so did not test individual Chart.  Anyway this is now fixed in latest push.

tturnerswdev33:
Everything is complete except moving TableLink to its own component or to Cove Media Controls.

adamdoe (approved):

PR #582, opened 2023-02-14
Feature/dev 2949 map hover updates

tturnerswdev33:
Removed weird vite config file

Move "Hide Geography Column Name In Tooltip" option from the Visual area of the editor to under Geography dropdown in Columns area of the editor
"HIDE PRIMARY COLUMN NAME IN TOOLTIP" does not appear to be having any effect. Please check what it is attempting to do, and discuss that will Bill to see if it is still needed.
If it is still needed move "HIDE PRIMARY COLUMN NAME IN TOOLTIP" option under Geography dropdown in Columns
Add a new text box option under the above options under the Geography dropdown. Label should be "Geography Label Override". Should default to blank, meaning the map would use the default Geography Label (e.g. State: Utah or County: Broome depending on the map). If it is populated, the Geography Label will use the override (e.g. <override value>: Utah>

- [x] Standalone Component
- [ ] Standalone Full Editor
- [ ] CDC Internal Checks


packages/editor/vite.config.js.timestamp-1675717859489.mjs
```diff
@@ -0,0 +1,58 @@
+// ../../generateViteConfig.js
```
adamdoe, line 1: fix: Please see if this is needed and remove if possible.

packages/map/src/CdcMap.jsx
```diff
@@ -1277,6 +1279,13 @@ const CdcMap = ({ className, config, navigationHandler: customNavigationHandler,
     }
   }, [state.dataTable])
 
+    // When geo label override changes
```
adamdoe, line 1282: nit: subjective, but I'd recommend keeping comments together above the useEffect instead of breaking them between lines.

packages/editor/vite.config.js.timestamp-1675717859489.mjs
tturnerswdev33: removed

packages/map/src/CdcMap.jsx
tturnerswdev33: OK changed it but I agree it really is a nit :-)

PR #600, opened 2023-02-23
Bug/dev 3163 charts support lg sm nums

tturnerswdev33:
( well that was short lived as it did not get merged before other PRs... we can revisit it on some other branch after the next release and get the eslint errors out though.... **FIXED ALL ESLINT ERRORS IN THE PROJECT!!!!!!!!!!!!**)

If "Round to Decimal points" setting is blank then roundTo is undefined and on load added code to SKIP rounding.  The effect on other data sets is unknown but this is needed to fix 3163 edge case.

For small numbers with decimals, the rounding still occurs if roundTo set to 0 but it leads to nonsensical X Axis values.  The fix in this code is to limit the number of tickmarks to not exceed the MAX of the data.
There is also a slight change to that if there are negative numbers which means min < 0

Update: Fixed issue when Num of Ticks set to Auto and also another issue with it not working on Horizontal charts.

- [x] Standalone Component
- [x] Standalone Full Editor
- [ ] CDC Internal Checks


packages/chart/src/components/LinearChart.jsx
```diff
@@ -209,13 +212,18 @@ export default function LinearChart() {
   }
 
   const countNumOfTicks = axis => {
+
```
adamdoe, line 215: nit: large spacing

packages/chart/src/components/LinearChart.jsx
```diff
@@ -209,13 +212,18 @@ export default function LinearChart() {
...
+    
     // function get number of ticks based on bar type & users value
     const isHorizontal = config.orientation === 'horizontal'
     const { numTicks } = config.runtime[axis]
     let tickCount = undefined
 
     if (axis === 'yAxis') {
       tickCount = isHorizontal && !numTicks ? data.length : isHorizontal && numTicks ? numTicks : !isHorizontal && !numTicks ? undefined : !isHorizontal && numTicks && numTicks
+      // DEV 3163 - to fix edge case of small numbers with decimals
+      if (tickCount > max) tickCount = min < 0 ? Math.round(max) * 2 : Math.round(max); // cap it and round it its not an integer
```
adamdoe, line 226: nit: readability

packages/chart/src/components/LineChart.jsx
```diff
@@ -18,6 +18,8 @@ export default function LineChart({ xScale, yScale, getXAxisData, getYAxisData,
   const cleanedData = cleanData(data, config.xAxis.dataKey)
   const { yScaleRight } = useRightAxis({ config, yMax, data, updateConfig })
 
+  //console.log("LineChart ###CleanedData=", cleanedData)
```
adamdoe, line 21: nit: remove comments unless they should stay in.

adamdoe (changes requested):
fix: add commas no longer appears to be working on the axis

packages/chart/src/components/LinearChart.jsx
tturnerswdev33: It's just a ternary... not sure what you want here.
I can do this instead of all on one line... is this what you want?  
      if (tickCount > max) {
          // cap it and round it its not an integer
          tickCount = min < 0 ? Math.round(max) * 2 : Math.round(max); 
      }

packages/chart/src/components/LineChart.jsx
tturnerswdev33: Removed it.  No just accidentally left in.

packages/chart/src/components/LinearChart.jsx
adamdoe: What you have is fine. I know it's a ternary. I was just pointing out that I think the cognitive load, at least for me, felt moderate. Maybe a small function or additional comments would help resolve that. It might be fine when you're actually in the code.

adamdoe (changes requested):
fix: When I opened the first sample file the comma support wasn't working. 

When I ran eslint I saw minor updates we could add for CdcChart and LinearChart (removing unused variables). I also see ~10 files that aren't formatted with prettier. Consider running `prettier --write "./packages/chart/src/**/*.jsx"` if you skip the eslint check.

tturnerswdev33:
Spent hours on this trying to figure out why line chart is not drawing...  my last conclusion was that something getting merged in from test is breaking it when I thought it was something in my edits.
I may still need to revisit fixing commas and maybe even fixing the small vs big data... dont know without getting line charts working on this branch.

.prettierrc.json
adamdoe, line 1: Remove this file

packages/chart/src/components/Filters.jsx
```diff
@@ -38,6 +38,7 @@ const useFilters = () => {
     // reset to first item in values array.
     newFilters.map(filter => {
       filter.active = filter.values[0]
+      return null
```
adamdoe, line 41: Did you test this? I know we've run into issues with `forEach` causing issues before with eslint but I wonder if we should be returning filter instead of null here.

packages/chart/src/components/Legend.jsx
```diff
@@ -25,12 +25,14 @@ const Legend = () => {
       let tmp = {}
       colsToKeep.map(col => {
         tmp[col] = isNaN(dataItem[col]) ? dataItem[col] : dataItem[col]
+        return null
```
adamdoe, line 28: similar as previous comments

packages/chart/src/CdcChart.jsx
```diff
@@ -658,8 +658,14 @@ export default function CdcChart({ configUrl, config: configObj, isEditor = fals
     // Use commas also updates bars and the data table
     // We can't use commas when we're formatting the dataFormatted number
     // Example: commas -> 12,000; abbreviated -> 12k (correct); abbreviated & commas -> 12 (incorrect)
+    //
+    // DEV 3163 edge case for small numbers with decimals
+    // - if roundTo undefined which means it is blank, then do not round
     if ((axis === 'left' && commas && abbreviated) || (axis === 'bottom' && commas && abbreviated)) {
       num = num // eslint-disable-line
+      //   } else if (roundTo === undefined) {
```
adamdoe, line 666: fix: remove commented code

packages/chart/src/components/LinearChart.jsx
```diff
@@ -91,6 +100,18 @@ export default function LinearChart() {
       }
     }
 
+    // DEV-3263 - If Confidence Intervals in data, then need to account for increased height in max for YScale
```
adamdoe, line 103: Personally adding the ticket numbers in feels confusing. This is 3163 but the comment references 3263. It could be a spelling error or it could be two different tickets. I still feel like ticket numbers can be left to git.

packages/chart/src/components/PieChart.jsx
```diff
@@ -2,7 +2,7 @@ import React, { useContext, useState, useEffect, useRef } from 'react'
 import { animated, useTransition, interpolate } from 'react-spring'
 import { Tooltip as ReactTooltip } from 'react-tooltip'
 
-import Pie, { ProvidedProps, PieArcDatum } from '@visx/shape/lib/shapes/Pie'
+import Pie from '@visx/shape/lib/shapes/Pie'
```
adamdoe, line 5: Cleanup, good.

packages/chart/src/hooks/useReduceData.js
```diff
@@ -57,13 +57,15 @@ function useReduceData(config, data) {
...
     }
+    //console.log("NOTE if you comment this out line chart does not work - max",max)
     return max
   }
 
   const getMinValueFromData = () => {
     let min
     const minNumberFromData = Math.min(...data.map(d => Math.min(...config.runtime.seriesKeys.map(key => (isNumber(d[key]) ? Number(cleanChars(d[key])) : 1000000000)))))
     min = String(minNumberFromData)
+    //console.log("min",min)
```
adamdoe, line 68: Nothing terrible, just remember to do a pass on diffs before pushing.

packages/chart/src/components/LinearChart.jsx
```diff
@@ -91,6 +100,18 @@ export default function LinearChart() {
...
+    if (config.visualizationType === 'Bar' || config.visualizationType === 'Combo') {
+      let ciYMax = 0
+      if (config.hasOwnProperty('confidenceKeys')) {
+        let upperCIValues = data.map(function (d) {
```
adamdoe, line 107: You don't need to change anything, but arrow functions can help with writing concise code and provide implicit returns:

 `data.map( (d) => d[config.confidenceKeys.upper])`

tturnerswdev33:
The return was blank returning nothing.  But eslint barked and 
we had previously tested returning a null as an equivalent.

packages/chart/src/components/LineChart.jsx
adamdoe: Removing now, looks like it snuck back in.

adamdoe (changes requested):
fix: seeing duplicated tick values in the full editor when using datasets. I'm also seeing them in the standalone editor when removing the hardcoded number of ticks values.

packages/chart/src/components/SparkLine.jsx
```diff
@@ -15,7 +15,7 @@ import useReduceData from '../hooks/useReduceData'
 import ConfigContext from '../ConfigContext'
 
 export default function SparkLine({ width: parentWidth, height: parentHeight }) {
-  const { transformedData: data, dimensions, config, parseDate, formatDate, currentViewport, seriesHighlight, formatNumber, colorScale, isNumber, handleChartAriaLabels } = useContext(ConfigContext)
+  const { transformedData: data, dimensions, config, parseDate, formatDate, currentViewport, seriesHighlight, formatNumber, colorScale, isNumber, handleChartAriaLabels } = useContext(ConfigContext) // eslint-disable-line
```
adamdoe, line 18: fix: don't use eslint here, remove the unused variables

packages/chart/src/components/LinearChart.jsx
```diff
@@ -25,7 +25,13 @@ import { DeviationBar } from './DeviationBar'
 
 // TODO: Move scaling functions into hooks to manage complexity
 export default function LinearChart() {
-  const { transformedData: data, dimensions, config, parseDate, formatDate, currentViewport, formatNumber, handleChartAriaLabels, updateConfig } = useContext(ConfigContext)
+  const { transformedData: data, dimensions, config, parseDate, formatDate, currentViewport, formatNumber, handleChartAriaLabels, cleanData, updateConfig } = useContext(ConfigContext)
+  // Just do this once up front otherwise we end up
+  // calling clean several times on same set of data (TT)
+  const cleanedData = cleanData(data, config.xAxis.dataKey)
```
adamdoe, line 31: fix: cleanedData is showing as unused for me. Can you verify this should be here?

packages/chart/src/components/LinearChart.jsx
```diff
@@ -25,7 +25,13 @@ import { DeviationBar } from './DeviationBar'
...
+
+  //const { transformedData: data, dimensions, config, parseDate, formatDate, currentViewport, formatNumber, handleChartAriaLabels, updateConfig } = useContext(ConfigContext)
```
adamdoe, line 33: fix: remove one of the useContext items

packages/chart/src/components/Filters.jsx
```diff
@@ -3,7 +3,7 @@ import ConfigContext from './../ConfigContext'
 import Button from '@cdc/core/components/elements/Button'
 
 const useFilters = () => {
-  const { config, setConfig, filteredData, setFilteredData, excludedData, filterData, runtimeFilters } = useContext(ConfigContext)
+  const { config, setConfig, filteredData, setFilteredData, excludedData, filterData, runtimeFilters } = useContext(ConfigContext) // eslint-disable-line
```
adamdoe, line 6: fix: don't use eslint ignore here, remove unused variables

packages/chart/src/components/SparkLine.jsx
```diff
@@ -96,6 +96,7 @@ export default function SparkLine({ width: parentWidth, height: parentHeight })
 
       yScale = config.runtime.xAxis.type === 'date' ? scaleLinear({ domain: [Math.min(...xAxisDataMapped), Math.max(...xAxisDataMapped)] }) : scalePoint({ domain: xAxisDataMapped, padding: 0.5 })
 
+      // eslint-disable-next-line
```
adamdoe, line 99: question: im not seeing any lint errors when removing this disable line

packages/chart/src/components/PieChart.jsx
```diff
@@ -31,21 +31,22 @@ export default function PieChart() {
...
   // Make sure the chart is visible if in the editor
+  /* eslint-disable react-hooks/exhaustive-deps */
   useEffect(() => {
     const element = document.querySelector('.isEditor')
     if (element) {
       // parent element is visible
       setAnimatePie(prevState => true)
     }
-  })
+  }) // eslint-disable-line
```
adamdoe, line 41: question: I don't believe this addition does anything.

packages/chart/src/components/Legend.jsx
```diff
@@ -170,13 +174,25 @@ const Legend = () => {
         </LegendOrdinal>
       </aside>
     ) : (
-      <aside id='legend' className={containerClasses.join(' ')} role='region' aria-label='legend' tabIndex={0}>
+      <aside
+        id='legend'
+        className={containerClasses.join(' ')}
+        role='region'
+        aria-label='legend'
+        tabIndex={0} // eslint-disable-line
```
adamdoe, line 182: fix: remove this line in favor of the error since we should address it for accessibility. Don't mute accessibility issues.

packages/chart/src/components/Legend.jsx
```diff
@@ -170,13 +174,25 @@ const Legend = () => {
...
     )
   return (
     config.visualizationType !== 'Box Plot' && (
-      <aside id='legend' className={containerClasses.join(' ')} role='region' aria-label='legend' tabIndex={0}>
+      <aside
+        id='legend'
+        className={containerClasses.join(' ')}
+        role='region'
+        aria-label='legend'
+        tabIndex={0} // eslint-disable-line
```
adamdoe, line 194: fix: remove this mute since we need to address the accessibility issue.

adamdoe (changes requested):
Please address the remaining feedback. In addition there are a number of `return null` statements on `.map()` functions. Lets return the actual item that needs to be returned in those statements.

adamdoe (changes requested):
Please address the remaining feedback. In addition there are a number of `return null` statements on `.map()` functions. Lets return the actual item that needs to be returned in those statements.

adamdoe (changes requested):
fix: In order to fix the large tooltip icons it looks like you should be able to play with the display property. Try `display: block` or `display: flex` with other styles to see if you can reduce the icon size. It looks like you should be able to reference other icons in other panels in the editor.

tturnerswdev33:
> fix: In order to fix the large tooltip icons it looks like you should be able to play with the display property. Try `display: block` or `display: flex` with other styles to see if you can reduce the icon size. It looks like you should be able to reference other icons in other panels in the editor.

I did try to change that between block, inline, flex etc. and it would not fix the problem.  I tried everything I could think of.  The only thing I could do was set a width and height but then I think that may mess up the text next to it.  Very baffling as the code looks the same as the code in single component in same location yet does not work in editor.  (FYI THIS HAS BEEN BROKEN FOR AWHILE AND NOT SOMETHING I BROKE)

packages/chart/src/components/Legend.jsx
tturnerswdev33: Sure I will remove it but whoever you said for me to check with she said just leave it there. i.e. leave tabIndex alone

tturnerswdev33:
> fix: seeing duplicated tick values in the full editor when using datasets. I'm also seeing them in the standalone editor when removing the hardcoded number of ticks values.

This has been fixed.

packages/chart/src/components/SparkLine.jsx
tturnerswdev33: OK done

packages/chart/src/components/LinearChart.jsx
tturnerswdev33: removed

packages/chart/src/components/Filters.jsx
tturnerswdev33: Fixed

packages/chart/src/components/SparkLine.jsx
tturnerswdev33: removed

packages/chart/src/components/Legend.jsx
tturnerswdev33: removed

PR #602, opened 2023-02-24
DEV-3219 Added padding to LinearChart max value so that Y axis always above th…

tturnerswdev33:
Added padding to calculation of "max" for YScale so that data never exceeds the scale on the Y Axis.
Works for any charts stemming from LinearChart

Per discussion added "Add Padding to Value Axis Scale" checkbox on EditorPanel for chart under "LEFT VALUE AXIS" to enable or disable the padding function. 

- [x] Standalone Component
- [x] Standalone Full Editor
- [ ] CDC Internal Checks


tturnerswdev33:
The JSON files are for testing.  They could be removed if you want.

mpallansch (approved):

tturnerswdev33:
Hey @mpallansch , just today (a few days after working on this) I see lines 117-122 in LinearChart and gitLens says you added that like 2 years ago.  It seems to be similar to this ticket.  Maybe your code doesn't work if they don't set the padding in config???
If so we could just set a default and then perhaps my code is not needed....

PR #613, opened 2023-03-02
Dev 784 show special classes in legend last

tturnerswdev33:
This applies to all MAPS. 

Added checkbox under Legend for "Show Special Classes Last" that when checked moves the special classes under Columns to the bottom of the legend. It was tricky because the component lookup to get the state and tooltip colors was using an index into a legend array but that array needed to have items moved to the end which in turn messed up the color lookup because the array rows are rearranged.  Solution was to add a new prop to the legend rows array for "bin" meaning data bin that color is assigned to.  Then when doing color look up you have to find the row where the bin number matches the lookup index.

Last push fixes ALL of the eslint errors in CdcMap.js

- [x] Standalone Component
- [ ] Standalone Full Editor
- [ ] CDC Internal Checks


tturnerswdev33:
Tomorrow FRIDAY I will go through the eslint errors and warnings.  Did not have time to do that today.

packages/map/src/CdcMap.jsx
```diff
@@ -177,8 +177,8 @@ const CdcMap = ({ className, config, navigationHandler: customNavigationHandler,
 
   // Tag each row with a UID. Helps with filtering/placing geos. Not enumerable so doesn't show up in loops/console logs except when directly addressed ex row.uid
   // We are mutating state in place here (depending on where called) - but it's okay, this isn't used for rerender
+  // eslint-disable-next-line
   const addUIDs = useCallback((obj, fromColumn) => {
-    // eslint-disable-line
```
adamdoe: @tturnerswdev33 - I updated this. `// eslint-disable-line` here, the way you had it, would do nothing. Added the eslint ignore above the function as `// eslint-disable-next-line`

packages/map/src/CdcMap.jsx
```diff
@@ -254,10 +254,10 @@ const CdcMap = ({ className, config, navigationHandler: customNavigationHandler,
     obj.data.fromColumn = fromColumn
   })
 
+  // eslint-disable-next-line
   const generateRuntimeLegend = useCallback((obj, runtimeData, hash) => {
-    // eslint-disable-line
     const newLegendMemo = new Map() // Reset memoization
-    var primaryCol = obj.columns.primary.name,
+    let primaryCol = obj.columns.primary.name,
```
adamdoe, line 260: @tturnerswdev33 - removed var in favor of let here.

packages/map/src/CdcMap.jsx
```diff
@@ -582,7 +582,8 @@ const CdcMap = ({ className, config, navigationHandler: customNavigationHandler,
           breaks.unshift(0)
         }
 
-        breaks.foreach((item, index) => {
```
adamdoe: @tturnerswdev33 - remember to test. Changing this to foreach broke major functionality on maps.

adamdoe (approved):
@tturnerswdev33 - The functionality looks good in standalone and the full editor now. 

A few things I saw that I resolved in a separate commit:

- major breaking change from `map` to `forEach` was reset, **remember to test changes**.
- `// eslint-ignore-line` should be unsed to ignore inline code, `// eslint-ignore-next-line` should be used to ignore code below the line. You might have fixed this in another commit.
- I made a small change from `var` to `let`. I'm not sure why it was changed from the original `const`, but set it to `let` to be compatible with any new changes you added.
- In `.filter() or .map() functions` try to use descriptive argument names on the arrow functions. I'm guilty of this, so I let it go. It wasn't immediately obvious what `o` was on one of the functions you added. Maybe it's an object, or maybe it's something else. It doesn't matter for this PR - just something to keep in mind while coding.

PR #619, opened 2023-03-07
Dev 3303 fix special classes bugs

tturnerswdev33:
This ended up 2 phase ticket but first part needed fixing to narrow down 2nd part.
Part 1). Old map config files with Special Classes set to strings.  Need code to repair that and convert to object with key, value, label.  There was code to do that in EditorPanel but it was in wrong place AND was not fixing it in the state so essentially was not fixing these cases.
Part 2). If you set a Special Class for say Rate and give it label of LEMONS, then LEMONS would show up anywhere the rate value matched ANY column key in both hover and data table.  The label should only show if the column key MATCHES the selected special class key.  Had to fix this in CdcMap and in DataTable but same fix both places.

- [x] Standalone Component
- [ ] Standalone Full Editor
- [ ] CDC Internal Checks


adamdoe (approved):
@tturnerswdev33 - functionality tested in standalone and full editor from what I can see. Nice work! I'm approving, but a few things to look out for in the future:

- fixed: You committed an empty file titled `package alias`, remember to take your time and review your commits when you push them. Try to set aside some time to open vs codes git tools or use the command line to see what you pushing.

- Code commenting.
    - nit: Avoid all-caps comments as it stands out. 
    - nit: There's no need to include the ticket number above chunks of codes. We have a git hook that formats commits as `[DEV-1234]: your commit message` if your concerned about tying the piece of code to a ticket. See: `prepare-commit-message` file for additional setup instructions if your interested.
    - nit: Lengthy comments should be kept to a minimum. Again, subjective, and some things need extra explanation. If you feel it's warranted please leave them in. However, using more expressive variable and function names or breaking up large functions into smaller, more manageable ones is what I would personally aim for.

PR #783, opened 2023-08-10
[cleanup] 4.23.8: brush ticket improvements

adamdoe:


tturnerswdev33:
Regarding the cleanup edits:

Removed unnecessary imports that were added while adding visx/brush. I'm guessing these were added while experimenting.
- yes all from the original example file in visx library and I guess some were not needed

Removed what appeared to be an unused function initBrushData
- originally I was experimenting with different data to start off the brush and ultimately kept editing and trimming it, so yes in the end useState(data) was the final start of the brush data rather than useState(initBrushData)

There was a lot of unused code near the top that looked like it was copied in, possibly from a visx example page, that wasn't used. I went ahead and removed that, ie. near PATTERN_ID.
- same as above... from original example in visx lib

I moved the remaining settings that were actually being used to a prop on the initial state. For example, the pattern id and line colors. Since these are essentially settings we can manipulate, we my as well put them on the state to keep the file clean.
- sure that makes sense

I moved the brush handle out into its own component for re-usability purposes and to keep the file smaller.
Minor deconstruction on showChartBrush. This should arguable be moved to something like config.brush.showChartBrush but I left it where it is for now.
- ok go ti

tturnerswdev33:
All of this looks fine to me.

tturnerswdev33 (commented):
I left all the console.logs in there bc there is much more work to do on this when getting it to work on bar charts.

There are also issues with the responsive tick marks that Murad is working on unless he is finished, as well as weird number of tick marks from visx which I dont think we can control.

PR #798, opened 2023-08-23
Feature/logi to wcms poc

joshlacey:

- [ ] Standalone Component
- [ ] Standalone Full Editor
- [ ] CDC Internal Checks


mpallansch (commented):
Looks good! My only immediate notes would be that the final PR should target the test branch instead of the main branch, as these changes will be merged in with other developer changes that are going out in the next release. Doe, Adam (CDC/IOD/OC) (CTR) do you know if any other developers are working on the dashboard package for the next release? If so, it might be better to revert the typescript, as otherwise the file names being changed might prevent an easy merge.

 

Also, I did notice that the URL filters apply now to all datasets that use URLs. That should be fine for now, but we may need to bring it up with Bill, as there may be some dashboards that require multiple dataset pulled from API, that don't share QS params.

packages/dashboard/src/CdcDashboard.tsx
```diff
@@ -37,11 +37,100 @@ import './scss/main.scss'
 import '@cdc/core/styles/v2/main.scss'
 import AdvancedEditor from '@cdc/core/components/AdvancedEditor'
 
+export type APIFilter = {
+  apiEndpoint: string
+  valueSelector: string
+  textSelector: string
+}
+
+export type SharedFilter = {
```
joshlacey, line 46: @adamdoe I think moving types out which are shared like this one makes sense. This one should go maybe into a /core/types folder as this could be used by any of the packages.

packages/dashboard/src/CdcDashboard.tsx
```diff
@@ -37,11 +37,100 @@ import './scss/main.scss'
...
+  datasets: DataSets
+  dataFileName: string
+  table: any
+  rows: any[]
+  data: any
+  filterBehavior: string
+}
+
+type APIFilterDropdowns = {
+  // null means still loading
```
joshlacey, line 117: this type is not exported and is not relevant to any other component than CdcDashboard. I think this one should remain in this file.

packages/dashboard/src/CdcDashboard.tsx
```diff
@@ -100,21 +189,22 @@ const VisualizationsPanel = ({ loadConfig, config }) => (
...
+    <AdvancedEditor loadConfig={loadConfig} state={config} convertStateToConfig={undefined} />
   </div>
 )
 
 export default function CdcDashboard({ configUrl = '', config: configObj = undefined, isEditor = false, isDebug = false, setConfig: setParentConfig }) {
-  const [config, setConfig] = useState(configObj ?? {})
-  const [data, setData] = useState([])
-  const [filteredData, setFilteredData] = useState()
-  const [loading, setLoading] = useState(true)
+  const [config, setConfig] = useState<Config | null>(configObj ?? null)
```
joshlacey, line 197: @mpallansch @adamdoe setConfig is defined twice... as a prop and as a setState function...

adamdoe (approved):

adamdoe:
I'm less familiar with the data fetching inside of dashboards, so I'll defer the final approval to @mpallansch . If there's anything we need to discuss further let me know!

packages/dashboard/src/CdcDashboard.tsx
```diff
@@ -219,14 +310,15 @@ export default function CdcDashboard({ configUrl = '', config: configObj = undef
   }
 
   const loadConfig = async () => {
-    let response = configObj || (await (await fetch(configUrl)).json())
+    setLoading(true)
+    let response: Config = configObj || (await (await fetch(configUrl)).json())
     let newConfig = { ...defaults, ...response }
-    let datasets = {}
```
joshlacey, line 224: @adamdoe here datasets is an Object, setData is later called and the object is passed.

packages/dashboard/src/CdcDashboard.tsx
```diff
@@ -187,8 +278,8 @@ export default function CdcDashboard({ configUrl = '', config: configObj = undef
             }
           }
 
-          newDatasets[datasetKeys[i]].runtimeDataUrl = dataUrlFinal
-          newData[datasetKeys[i]] = newDataset
+          newDatasets[datasetKey].runtimeDataUrl = dataUrlFinal
+          newData[datasetKey] = newDataset
           datasetsNeedsUpdate = true
         }
       }
```
joshlacey, line 285: the setData function below is again setting a data object

PR #846, opened 2023-09-26
Feature/logi to wcms poc filter layout

joshlacey:

- [ ] Standalone Component
- [ ] Standalone Full Editor
- [ ] CDC Internal Checks


joshlacey:
@mpallansch @adamdoe Can we do a Squash merge on this? We should start doing that going forward to remove unnecessary commit messages and simplify merging/rebasing.

packages/dashboard/src/types/Config.ts
```diff
@@ -0,0 +1,44 @@
+import { DataSet } from "./DataSet"
```
adamdoe, line 1: Incorrect import

packages/dashboard/src/CdcDashboard.tsx
```diff
@@ -29,170 +29,27 @@ import CdcFilteredText from '@cdc/filtered-text'
 import Grid from './components/Grid'
 import Header, { FilterBehavior } from './components/Header'
 import defaults from './data/initial-state'
-import Widget from './components/Widget'
 import DataTable from './components/DataTable'
 import MediaControls from '@cdc/core/components/MediaControls'
 
 import './scss/main.scss'
 import '@cdc/core/styles/v2/main.scss'
```
adamdoe, line 36: It looks like you removed AdvancedEditor here. While the apply button doesn't work, I like having this in if I need to copy a specific example of a Dashboard to my local. Can we keep this in?

packages/core/helpers/DataTransform.ts
```diff
@@ -11,7 +14,7 @@ export class DataTransform {
 
   //Performs standardizations that can be completed automatically without use input
   autoStandardize(data) {
-    const errorsFound = []
+    const errorsFound: any[] = []
 
     // Empty data
     if (0 === data.length) {
```
adamdoe, line 20: I'm receiving an error at this point where data doesn't have a length. Consider using a short circuit ternary or checking for data before this point.

adamdoe (changes requested):
The url helpers are a nice edition in core and I appreciate the reusability and structure added. One thing that might make our helpers  better is to create more general files, ie. number.js, date.js, and in this case `url.js` to create files that can provide more generic helpers and more than just a single returned function (ie. combining getFileExtension and getQueryParams). Nothing needed at this time on that comment.   

Nice work adding the TS in and also adding try/catch blocks!

Removed AdvancedEditor, is that intentional?

Bad import in `packages/dashboard/src/types/Config` on `Dataset` not `DataSet`  Unable to create a dashboard after re-building.

Blocking error on DataTransform for the default file. Double check any issues there.

adamdoe (commented):
The url helpers are a nice edition in core and I appreciate the reusability and structure added. One thing that might make our helpers  better is to create more general files, ie. number.js, date.js, and in this case `url.js` to create files that can provide more generic helpers and more than just a single returned function (ie. combining getFileExtension and getQueryParams). Nothing needed at this time on that comment.   

Nice work adding the TS in and also adding try/catch blocks!

Removed AdvancedEditor, is that intentional?

Bad import in `packages/dashboard/src/types/Config` on `Dataset` not `DataSet`  Unable to create a dashboard after re-building.

Blocking error on DataTransform for the default file. Double check any issues there.

packages/dashboard/src/CdcDashboard.tsx
```diff
@@ -720,7 +782,8 @@ export default function CdcDashboard({ configUrl = '', config: configObj = undef
             </div>
           )
         })}
-        {config.filterBehavior === FilterBehavior.Apply && <button onClick={applyFilters}>GO!</button>}
+
+        {!isAutoLoadRow && <button onClick={applyFilters}>GO!</button>}
```
mpallansch, line 662: The default for already saved configs will be not not have the FilterBeahvior.Apply set, however this will still show the go button in that case. Need to rework logic to hide go button when FilterBehavior is unset.

packages/dashboard/src/CdcDashboard.tsx
```diff
@@ -251,9 +160,13 @@ export default function CdcDashboard({ configUrl = '', config: configObj = undef
 
   const getVizKeys = (_config): string[] => Object.keys(config?.visualizations || {})
```
mpallansch, line 120: Not sure where this function was added, but the config reference needs to be the local reference (_config) instead of the global one:

  const getVizKeys = (_config): string[] => Object.keys(_config?.visualizations || {})

mpallansch (changes requested):
Additionally, I'm not sure where the "generateValuesForFilter" function in CdcDashboard.tsx was refactored but it does not appear to be working with the current data structure:

{
  dataKey1: [{...}, {...}, ...],
  dataKey2: [{...}, {...}, ...]
}

This will apply to data filters (non URL filters). The function on the integration branch is working, you can look at that to see what might be wrong with the refactor.

You can add showDropdown: true to the filter in this example file default-filter-control.json to see where the filter is not being populated.

packages/dashboard/src/CdcDashboard.tsx
joshlacey: I moved the VisualizationsPanel out to its own file as it's a complete component. AdvancedEditor was the only used in there so it was removed from the imports here as Dashboard doesn't use it.

packages/core/helpers/DataTransform.ts
joshlacey: I tried using the default example ` <div class="react-container" data-config="/examples/default.json"></div>` I'm not seeing any error here though.

adamdoe:
@joshlacey - we had hotfixes go out for this release so you'll need to rebase here.

packages/dashboard/src/CdcDashboard.tsx
```diff
@@ -639,51 +538,90 @@ export default function CdcDashboard({ configUrl = '', config: configObj = undef
...
+        // setData to empty object because we no longer have a data state.
+        setData({})
+        setFilteredData({})
+      }
+    }
+  }
+
+  const Filters = ({ hide, autoLoad }: { hide?: number[]; autoLoad?: boolean }) => {
+    if (!config) return <></>
+    const isLegacyFilter = !config.filterBehavior
```
joshlacey, line 607: @mpallansch will this resolve the legacy filter issue?

packages/dashboard/src/CdcDashboard.tsx
joshlacey: @mpallansch I added an isLegacyFilter variable which may resolve this? can you review that and let me know? I added a comment next to the variable

joshlacey:
> Additionally, I'm not sure where the "generateValuesForFilter" function in CdcDashboard.tsx was refactored but it does not appear to be working with the current data structure:
> 
> { dataKey1: [{...}, {...}, ...], dataKey2: [{...}, {...}, ...] }
> 
> This will apply to data filters (non URL filters). The function on the integration branch is working, you can look at that to see what might be wrong with the refactor.
> 
> You can add showDropdown: true to the filter in this example file default-filter-control.json to see where the filter is not being populated.

This change is already on the Test branch. Can I raise and issue with this and create a separate PR for that? @mpallansch @adamdoe 

I refactored this without thoroughly looking at the legacy data structure. I need to create separate functions I think for API filter data VS this structure.

mpallansch (approved):

adamdoe (approved):

PR #897, opened 2023-10-19
Feature/dev 5806 grouped table

joshlacey:

- [ ] Standalone Component
- [ ] Standalone Full Editor
- [ ] CDC Internal Checks


packages/core/components/DataTable/components/ChartRows.tsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  if (isVertical) {
+    return (
+      <>
+        {rows.map((row, index) => {
+          return (
+            <tr key={`${row}__${index}`} role='row'>
+              {dataSeriesColumns.map((column, j) => {
+                ;<td key={`${row}__${index}_${j}`} tabIndex={0} role='gridcell' id={`${runtimeData[config.runtime?.originalXAxis?.dataKey]}--${row}`}>
```
joshlacey, line 40: @adamdoe @mpallansch @Atash3000 does anyone know what the id is doing on this td cell? is there some popup or something that's referencing it? Can it be removed?

packages/map/src/CdcMap.tsx
```diff
@@ -1252,14 +1248,7 @@ const CdcMap = (props: CdcMapProperties) => {
                 vizTitle={general.title}
                 viewport={currentViewport}
                 formatLegendLocation={formatLegendLocation}
-                setFilteredCountryCode={setFilteredCountryCode}
```
joshlacey, line 1255: @adamdoe @Atash3000 @mpallansch I removed this. I'm not sure what it's doing, but I know that it's being called from an onClick handler on a <td> which I don't think is 508 compliant. I'll make a comment there as well

packages/core/components/DataTable.jsx
```diff
@@ -1,759 +0,0 @@
...
-                        specialValFound = config.legend.specialClasses[i].label
-                      }
-                    }
-                  }
-                }
-                cellValue = specialValFound ? displayDataAsText(specialValFound, column) : displayDataAsText(runtimeData[row][config.columns[column].name], column)
-              }
-
-              return (
-                <td tabIndex='0' role='gridcell' onClick={e => (config.general.type === 'bubble' && config.general.allowMapZoom && config.general.geoType === 'world' ? setFilteredCountryCode(row) : true)}>
```
joshlacey, line 239: pretty sure and onClick on a <td> isn't 508 compliant. can we handle this some other way? @adamdoe said the feature is broken anyway.

@mpallansch @Atash3000

tturnerswdev33:
Hey Josh,
I’m not working on this project any more.
Regards,
-Tim Turner

joshlacey:
> 
> Is the grouped functionality accessible through the editor? How would an end user create a grouped table?

I think there's some UX guidance that should be given regarding the editor, we aren't getting that right now, so every new feature we're just supposed to pick a place within the editor to add the functionality. We do have the option to manually edit the json object within the editor, I would consider this MVP until given direct UX guidance for where to put things within Editor. In my opinion, while it was fine previously for developers to decide where these controls should live, as a larger distributed team I don't know if we should keep having to make those decisions. I really want to make a point that UX needs to be looking at editor and making determinations on how it functions. @adamdoe @mpallansch what are your thoughts?

joshlacey:
> Was having trouble doing the review, it looks like maps are crashing when created in the editor. It looks like its due to a different refactor though.

do you mean that the issue is happening on dev branch?

mpallansch:
> > Was having trouble doing the review, it looks like maps are crashing when created in the editor. It looks like its due to a different refactor though.
> 
> do you mean that the issue is happening on dev branch?

I had checked out your branch for testing, but yes the issue is happening on dev branch as well.

mpallansch:
> > Is the grouped functionality accessible through the editor? How would an end user create a grouped table?
> 
> I think there's some UX guidance that should be given regarding the editor, we aren't getting that right now, so every new feature we're just supposed to pick a place within the editor to add the functionality. We do have the option to manually edit the json object within the editor, I would consider this MVP until given direct UX guidance for where to put things within Editor. In my opinion, while it was fine previously for developers to decide where these controls should live, as a larger distributed team I don't know if we should keep having to make those decisions. I really want to make a point that UX needs to be looking at editor and making determinations on how it functions. @adamdoe @mpallansch what are your thoughts?

If the question is where to add inputs within the editor, I think developers should be able to pick a place and things can easily be moved if UX/UAT have concerns about the location. In this instance, I think somewhere in the "Data Table" sections of Maps, Charts, and Dashboard would be the obvious place. If there is a greater question about the types of inputs needed for a feature, that could definitely wait for UX to weigh in. However I think its helpful in most cases to be able to test changing things within the editor to quickly iterate through different permutations of the functionality without having to make manual edits to the config files. Also, from what I can tell from the implementation, it looks like a single straightforward drodown of data columns allowing the user to pick which column to group the data table by.

joshlacey:
> > > Is the grouped functionality accessible through the editor? How would an end user create a grouped table?
> > 
> > 
> > I think there's some UX guidance that should be given regarding the editor, we aren't getting that right now, so every new feature we're just supposed to pick a place within the editor to add the functionality. We do have the option to manually edit the json object within the editor, I would consider this MVP until given direct UX guidance for where to put things within Editor. In my opinion, while it was fine previously for developers to decide where these controls should live, as a larger distributed team I don't know if we should keep having to make those decisions. I really want to make a point that UX needs to be looking at editor and making determinations on how it functions. @adamdoe @mpallansch what are your thoughts?
> 
> If the question is where to add inputs within the editor, I think developers should be able to pick a place and things can easily be moved if UX/UAT have concerns about the location. In this instance, I think somewhere in the "Data Table" sections of Maps, Charts, and Dashboard would be the obvious place. If there is a greater question about the types of inputs needed for a feature, that could definitely wait for UX to weigh in. However I think its helpful in most cases to be able to test changing things within the editor to quickly iterate through different permutations of the functionality without having to make manual edits to the config files. Also, from what I can tell from the implementation, it looks like a single straightforward drodown of data columns allowing the user to pick which column to group the data table by.

In an effort to keep this specific PR as small as possible I'd like to not make any changes to the editor component. I think UX still has a few comments about the styling for the datatable and I'd like to get that resolved before anyone in WCMS is able to use it. Not exposing it in the editor to me is kinda like having a feature flag disabled. We aren't ready for it to be used in production but it is in the code base. I'm going to have a later ticket to implement the page and I can make the editor capability to group the table part of that ticket. I'll also bring this conversation up in standup today so we can move forward. UX is only checking this thing in storybook as of right now so I think the editor isn't part of scope of the original ticket.

packages/map/src/CdcMap.tsx
joshlacey: @adamdoe and I were talking about this in chat and he said that the feature doesn't even seem to be working in prod. We weren't sure why this control was needed on the datatable either. I would suggest if someone wants to put it back, instead of trying to add a click handler to the outer `<td>` we should instead pass a button or anchor tag into the table that can be clicked. This should indicate to screen readers and regular users that an action is being taken with the click.

mpallansch (approved):

PR #900, opened 2023-10-20
[PORT-134] Markup Include TS & useReducer

adamdoe:
- Converts Markup Include to TypeScript
- Adds useReducer to Markup Include

### DIFF
--- a/packages/markup-include/src/CdcMarkupInclude.tsx
+++ b/packages/markup-include/src/CdcMarkupInclude.tsx
@@ -72,7 +72,6 @@ const CdcMarkupInclude = (props: CdcMarkupIncludeProps) => {
     const processedConfig = { ...(await coveUpdateWorker(response)) }
 
     updateConfig({ ...defaults, ...processedConfig })
-    // setLoading(false)
     dispatch({ type: 'SET_LOADING', payload: false })
   }, [])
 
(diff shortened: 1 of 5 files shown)


packages/markup-include/src/store/mi.actions.ts
```diff
@@ -0,0 +1,21 @@
...
+type Action<T, P> = { type: T; payload: P }
+
+type SET_CONFIG = Action<'SET_CONFIG', Config>
+type SET_LOADING = Action<'SET_LOADING', boolean>
+type SET_URL_MARKUP = Action<'SET_URL_MARKUP', string>
+type SET_MARKUP_ERROR = Action<'SET_MARKUP_ERROR', any>
+type SET_ERROR_MESSAGE = Action<'SET_ERROR_MESSAGE', any>
+type SET_COVE_LOADED_HAS_RAN = Action<'SET_COVE_LOADED_HAS_RAN', boolean>
+
+// prettier-ignore
```
joshlacey, line 12: can we set our linter line width to something lower? it only wraps after a really long line. I think this is fine for now, just making a comment for the future.

packages/markup-include/src/CdcMarkupInclude.tsx
```diff
@@ -61,7 +72,8 @@ const CdcMarkupInclude = ({ configUrl, config: configObj, isDashboard = false, i
     const processedConfig = { ...(await coveUpdateWorker(response)) }
 
     updateConfig({ ...defaults, ...processedConfig })
-    setLoading(false)
+    // setLoading(false)
```
joshlacey, line 75: remove comment

packages/markup-include/src/store/mi.actions.ts
adamdoe: Yes please!

PR #901, opened 2023-10-20
[PORT-134] Data Bite TS & useReducer

adamdoe:
- Move data bite to TS
- add useReducer


packages/data-bite/types/Config.ts
```diff
@@ -0,0 +1,39 @@
+export type Config = {
+  type: string
+  data: any[]
```
joshlacey, line 3: avoid using any where possible.

I think data shape is usually: Object[]

packages/data-bite/types/Config.ts
```diff
@@ -0,0 +1,39 @@
...
+  dataColumn: string
+  bitePosition: string
+  biteFontSize: number
+  fontSize: string
+  biteBody: string
+  imageData: {
+    display: string
+    url: string
+    alt: string
+    options: any[]
```
joshlacey, line 15: Object[] maybe?

packages/data-bite/types/Config.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+  dataFormat: {
+    roundToPlace: number
+    commas: boolean
+    prefix: string
+    suffix: string
+  }
+  biteStyle: string
+  filters: any[]
```
joshlacey, line 24: Object[] maybe?

PR #912, opened 2023-10-30
Feature/dev 4215a

Atash3000:
Added Brush chart to Line, Combo,Bars/regular/stacked/Vertical only , Area chart regular/stacked


packages/chart/src/CdcChart.tsx
```diff
@@ -1188,7 +1188,18 @@ export default function CdcChart({ configUrl, config: configObj, isEditor = fals
             {/* Visualization */}
             {config?.introText && config.visualizationType !== 'Spark Line' && <section className='introText'>{parse(config.introText)}</section>}
             <div
-              style={{ marginBottom: config.legend.position !== 'bottom' && config.orientation === 'horizontal' ? `${config.runtime.xAxis.size}px` : '0px' }}
+              style={{
```
adamdoe, line 1191: Simplify styles here using classes.

packages/chart/src/components/AreaChart.jsx
```diff
@@ -42,38 +36,19 @@ const AreaChart = props => {
...
-    // else dont return the other brush handles
+    return yScale(d[s.dataKey])
   }
 
   return (
     data && (
       <svg>
         <ErrorBoundary component='AreaChart'>
-          <Group className='area-chart' key='area-wrapper' left={Number(config.yAxis.size)} top={isBrush ? yMax * 1.3 : 0}>
+          <Group className='area-chart' key='area-wrapper' left={Number(config.yAxis.size)} top={0}>
```
adamdoe, line 51: I think we need to consider the spacing for the x axis label here. I'm getting overlapping when adjusting the x axis size with the x axis label and the brush slider. Similarily I'm seeing overlap with the ticks. We should try to account for some of the scenarios before merging.

packages/chart/src/components/ZoomBrush.jsx
```diff
@@ -0,0 +1,120 @@
+import { Brush } from '@visx/brush'
```
adamdoe, line 1: Since this is a new file can you convert some of this to TS?

packages/chart/src/scss/main.scss
```diff
@@ -14,6 +14,14 @@
 .flex-column-reverse {
   flex-direction: column-reverse;
 }
+.visx-brush {
+  & > rect {
+    stroke-width: 1px !important;
```
adamdoe, line 19: Do you need to use !important here?
It looks like rx isn't a valid css property, however, removing it breaks the look. Look into workarounds for this. It's also not supported by Firefox.

packages/chart/src/components/ZoomBrush.jsx
```diff
@@ -0,0 +1,120 @@
...
+        handleSize={8}
+        brushDirection='horizontal'
+        initialBrushPosition={initialPosition}
+        onChange={onBrushChange}
+      />
+    </Group>
+  )
+}
+
+function BrushHandle(props) {
```
adamdoe, line 99: Follow the arrow function component convention here.

adamdoe (changes requested):
At a glance this all looks wonderful. Nice job, Murad! I'm not sure if folks will have style feedback, but if you haven't gathered that yet, I'd reach out to our UX folks (ie. small dates on slider / general theming input). I added a few minor pieces of feedback here that should be addressed. Great work!

PR #925, opened 2023-11-07
[DEV-5189] Dashboard: filter charts with setBy key

adamdoe:
# DEV-5189 
- When dashboards have a vertical bar chart, and a setBy filter driven by that bar chart, update the colors of the bar chart.
- Add types on bar charts

### DIFF
--- a/packages/map/src/CdcMap.tsx
+++ b/packages/map/src/CdcMap.tsx
@@ -1129,7 +1129,7 @@ const CdcMap = (props: CdcMapProperties) => {
 
   if (!mapProps.data || !config.data) return <></>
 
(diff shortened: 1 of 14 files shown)


packages/map/src/CdcMap.tsx
```diff
@@ -1044,9 +1045,11 @@ const CdcMap = (props: CdcMapProperties) => {
     dispatch({ type: 'SET_RUNTIME_LEGEND', payload: legend })
   }, [runtimeData, config.legend.unified, config.legend.showSpecialClassesLast, config.legend.separateZero, config.general.equalNumberOptIn, config.legend.numberOfItems, config.legend.specialClasses]) // eslint-disable-line
 
+  console.log('dash', dashboardConfig)
```
joshlacey, line 1048: Can these be removed?

PR #934, opened 2023-11-14
Feature/dev 4523a

Atash3000:
Data Suppression


adamdoe (changes requested):
Use align-items center on the legend-items div to ensure the icon and text are centered properly. Also, check the alignment on the suppressed icon in relation to the bar. visx has helpers for vertical alignment and text alignment.

packages/core/components/LegendCircle.jsx
```diff
@@ -9,7 +9,8 @@ export default function LegendCircle({ fill, borderColor }) {
     height: '1em',
     width: '1em',
     border: borderColor ? `${borderColor}  1px solid` : 'rgba(0,0,0,.3) 1px solid',
-    backgroundColor: fill
+    backgroundColor: fill,
+    display: display
```
adamdoe, line 13: Remove duplicated display value above this one.

packages/core/helpers/DataTransform.ts
```diff
@@ -222,7 +222,7 @@ export class DataTransform {
     */
   cleanData(data: DataArray, excludeKey, testing = false): DataArray {
     let cleanedupData: DataArray = []
-    if(!Array.isArray(data)) debugger;
+    if (!Array.isArray(data)) debugger
```
adamdoe, line 225: Remove debugger statement

packages/chart/src/components/BarChart.Horizontal.tsx
```diff
@@ -14,10 +15,20 @@ import { type BarChartProps } from '../types/ChartProps'
 export const BarChartHorizontal = (props: BarChartProps) => {
   const { xScale, yScale, yMax, seriesScale } = props
   const { transformedData: data, colorScale, seriesHighlight, config, formatNumber, formatDate, parseDate, setSharedFilter, isNumber, getTextWidth, getYAxisData, getXAxisData } = useContext(ConfigContext)
-  const { isHorizontal, barBorderWidth, hasMultipleSeries, applyRadius, updateBars, assignColorsToValues, section, fontSize, isLabelBelowBar, displayNumbersOnBar, lollipopBarWidth, lollipopShapeSize, getHighlightedBarColorByValue, getHighlightedBarByValue } = useBarChart()
+  const { isHorizontal, barBorderWidth, hasMultipleSeries, applyRadius, updateBars, assignColorsToValues, section, fontSize, isLabelBelowBar, displayNumbersOnBar, lollipopBarWidth, lollipopShapeSize, getHighlightedBarColorByValue, getHighlightedBarByValue, generateIconSize } = useBarChart()
 
   const { HighLightedBarUtils } = useHighlightedBars(config)
-
+  const getIcon = (bar, barWidth) => {
```
adamdoe, line 21: If this function is a helper for all bar charts, you might consider putting it within the useBarChart hook as well.

adamdoe (changes requested):
I think this is looking good! 

However, I think we need to spend additional time on aligning the icons with the bars properly so we don't run into issues in the future. When I used the sample data and removed the filter to have five bars per date the icon was placed slightly to the right of the bar.

adamdoe (approved):

PR #951, opened 2023-11-28
[DEV-6202] Add Line Chart Circle Color Option

adamdoe:
- Added chart context type
- Added line chart props type
- Removed unused functions: handleAxisFormatting, formatDate in Line Chart
- Remvove unused variables: xAxis, yAxis values in Line Chart


Atash3000 (approved):
Works properly.

PR #962, opened 2023-12-01
Feature/826_data-viz-state-level-county-maps-type-of-navigation-map

petercolella:
* [DEV-826] update map type options for US-County Level geography subtype

image


petercolella:
@adamdoe this is ready for review.

adamdoe (approved):

PR #964, opened 2023-12-04
[DEV-6286] 4.23.11: updated logic for region highlighting

adamdoe:
@mpallansch & @Atash3000 - can you do a quick pass on this? It sets regions on grouped bar charts to account for the full bar width. Thanks!


Atash3000 (approved):
Works good!

PR #966, opened 2023-12-06
[combo-stacked-scaling] Resolve stacked combo scaling issue

adamdoe:


adamdoe:
@Atash3000 - can you do a quick review here?

PR #985, opened 2023-12-28
[dev-6596] data driven grouped tables

joshlacey:
added pie chart example and added grouped table example to storybook


packages/chart/src/components/EditorPanel.jsx
```diff
@@ -579,37 +456,6 @@ const EditorPanel = () => {
     updateConfig({ ...config }, newData)
   }
 
-  const removeSeries = seriesKey => {
```
adamdoe, line 582: Why is this removed?

packages/chart/src/components/EditorPanel.jsx
```diff
@@ -133,81 +85,6 @@ const Select = memo(({ label, value, options, fieldName, section = null, subsect
   )
 })
 
-const DataSuppression = memo(({ config, updateConfig, data }) => {
```
adamdoe, line 136: Why is this removed?

packages/chart/src/types/ChartConfig.ts
```diff
@@ -1,18 +1,7 @@
...
-    tooltips: boolean
-    prefix: string
-    suffix: string
-    forestPlot: boolean
-    startingPoint: string
-    forestPlotAlignRight: boolean
-    name?: string
-  }
-}
+export type ChartColumns = Record<string, Column>
```
adamdoe, line 4: Nice change, this helps me going forward!

adamdoe (commented):
Sorry, I didn't catch that this was just a draft. I think its looking good so far. I didn't dive into the table stuff deep, I'll trust that that's looking good since you're in the weeds with it. The two things that did come up for me were 1) deletions from EditorPanel. I'm wondering if they showed as unused but are actually needed. It seemed like a good chunk was stripped out and I'm pretty sure the DataSuppression component and remove series functions needs to stay in. So maybe double check deletions there. 2) The DataImport file was changed to use .json. Does that break csv imports?

packages/chart/src/components/EditorPanel.jsx
joshlacey: I saw it wasn't used. I can put it back if we need it. I just saw EditorPanel was 2k lines so wanted to cut where I could.

packages/chart/src/components/EditorPanel.jsx
joshlacey: I can put it back if we need.

joshlacey:
@adamdoe 
1) there was a data suppression section in the EditorPanel code that was commented out and I think that was the only place these were used. I'm usually of the opinion that it's better to remove unused code, it's alway present in the git history so if you need it in the future you can grab it then. I'm fine with putting it back though if you feel otherwise, doesn't affect this PR at all, was just doing some cleanup.

2) I'll need to double check about the csv import. I made the change because it was throwing an error because the url didn't end in .json, which is specific to Socrata. I'll need to make some change along these lines. I could remove this from this PR though as it doesn't affect the grouped tables. I was just trying to test external endpoints.

joshlacey:
@adamdoe some of the work to pull out the helper functions from CdcChart aren't needed for this PR. I started doing it because I needed to figure out something else, I think it simplifies CdcChart a lot though so I left it in. Let me know if you want me to put that change in a different PR.

Also I'm waiting for an okay from my DBA about the data structure for the table, once I get that I'll move this out of draft.

PR #999, opened 2024-01-04
Feature/6389_ticks-on-graph-look-thicker-in-some-VPs

petercolella:
- [DEV-6389] adds useLogScale boolean to ternary expression determining strokeWidth value


petercolella:
@adamdoe

It seems the stroke width ternary expression was first added in [this commit](https://github.com/CDCgov/cdc-open-viz/commit/1790c5898f9246f5bced1a8bd41d75352fb5cedc). I believe the change I've made maintains the original intent and fixes the bug, but I am unsure how to test it (that's why there's no screenshot). I can remove the strokeWidth prop entirely if you prefer. Thanks!

adamdoe (approved):
Looks good, thanks @petercolella!

PR #1017, opened 2024-01-11
[Dev-6648] add column to viz

joshlacey:
In order to add a column to the visualization I'll need to pivot two columns into one column. This will mean that I will have to override the colorScale that is passed from the CdcChart. I'll be handing this logic in the PieChart since adding the column is going to be specific to PieChart only. Once the instructions for a need for a pivot are passed to the PieChart, it will do the necessary steps to pivot the data and recreate the colorscale. This colorscale and updated data will then be passed to the Legend. Legend.Component will become a part of the PieChart component itself and Legend.tsx (the wrapper) will have logic to exclude rendering if the viz type is PieChart.


packages/chart/src/components/EditorPanel/EditorPanel.tsx
```diff
@@ -2642,6 +2642,18 @@ const EditorPanel = () => {
                                 </label>
                               )}
                             </li>
+                            <li>
```
adamdoe, line 2645: I think you might want to add visualizationType === 'Pie' here. It doesn't look like its used on any other chart types.

packages/core/components/DataTable/types/TableConfig.ts
```diff
@@ -2,43 +2,19 @@ import { Axis } from '@cdc/core/types/Axis'
...
-      outliers: string
-      values: string
-      total: string
-      lowerBounds: string
-      upperBounds: string
-    }
-    plots: []
-    categories: string[]
-  }
+  boxplot?: BoxPlot
```
adamdoe, line 14: Nice, work on TS throughout this pr!

adamdoe (changes requested):

PR #1023, opened 2024-01-16
Feature/dev 6513a

Atash3000:
Re fixed prelimary data for line charts and legend builded.


adamdoe:
@Atash3000 closing this PR since you had 6513b opened. If anything needs to be added from this PR let me know, but I think we're good.

PR #1033, opened 2024-01-20
Editor Context Provider

adamdoe:


joshlacey (commented):
overall looks good, I didn't see anything like this but if you're using a global panel from core folder I would suggest leaving it out at the EditorPanel.tsx level rather than nesting it under one of the Panel options.

PR #1035, opened 2024-01-22
Fixes Data Preview Not Showing

mpallansch:


adamdoe:
@mpallansch can we close this? It looks like the positioning is holding on test and changing to absolute positioning crops the gray area.

PR #1040, opened 2024-01-24
[dev-6597] Editor reducer refactor

joshlacey:
This change was necessary as a preliminary for the MultiDashboard work.


packages/core/types/Visualization.ts
```diff
@@ -2,20 +2,26 @@ export type Visualization = {
   autoLoad: boolean
   data: any
   dataDescription: Object
+  dataFileName: string
+  dataFileSourceType: string
   dataKey: string
+  datasets: any
```
adamdoe, line 8: I think this should be a record.

packages/core/types/Visualization.ts
```diff
@@ -2,20 +2,26 @@ export type Visualization = {
...
   editing: boolean
-  formattedData: any
+  formattedData?: Object[]
   general: any
   hide: any[]
+  multiDashboards?: any[]
   newViz: boolean
   openModal: boolean
   originalFormattedData: any
+  orientation: any
```
adamdoe, line 17: Likely 'vertical' | 'horizontal'

packages/core/types/Visualization.ts
```diff
@@ -2,20 +2,26 @@ export type Visualization = {
...
+  dataFileName: string
+  dataFileSourceType: string
   dataKey: string
+  datasets: any
   editing: boolean
-  formattedData: any
+  formattedData?: Object[]
   general: any
   hide: any[]
+  multiDashboards?: any[]
```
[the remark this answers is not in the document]
joshlacey: I'll fix this in the next PR. I forgot to remove it, was added prematurely.

packages/core/helpers/withDevTools.ts
```diff
@@ -0,0 +1,50 @@
...
+  send: (action: Action, newState: any) => any
+  unsubscribe: () => any
+  dispatch: (action: Action) => any
+  disconnect: () => any
+}
+
+// constants
+const withDevTools = typeof window !== 'undefined' && (window as any).__REDUX_DEVTOOLS_EXTENSION__
+const devTools: DevTools = !withDevTools ? null : (window as any).__REDUX_DEVTOOLS_EXTENSION__.connect()
+const devToolsStore = !withDevTools ? null : devTools
```
adamdoe, line 42: I think you might need to add an export here. I'm seeing the build fail without that. Can you add that in and re-verify that all actions are working? I'm seeing it working in redux dev tools now, but a second set of eyes would be nice.

joshlacey: good catch! I forgot I needed to export that. so basically as I have it designed right now it gets initialized by the Editor component. so only if you're running the editor you'll see the devtools stuff.

PR #1044, opened 2024-01-25
[DEV-6574] filter dashboard datasets

adamdoe:
There is a default example of the behavior in index.html. 

We are retrieving the parent filters from an API and then changing the child dataurl filters based on the value in the parent.


packages/dashboard/examples/filters/Alabama.json
joshlacey, line 1: Git won't even let me open this file because it's too large. can we reduce the file size or are these necessary?

joshlacey:
I was able to test that API Filters were still working. I think this looks good except for the large file sizes in the examples. I'd consider also using storybook for the example so we can have a clear demo of all of the different filter behaviors. I think you can use the fetchmock library to serve up your files?

PR #1049, opened 2024-01-26
Reworks Configuration Ready Logic

mpallansch:


packages/dashboard/src/components/Widget.tsx
```diff
@@ -266,14 +266,27 @@ const Widget = ({ data, addVisualization, type }: WidgetProps) => {
     }
   }, [data?.openModal])
 
+  let isConfigurationReady = false;
+  if(type === 'markup-include' && type === 'filter-dropdowns'){
```
joshlacey, line 270: this should be OR @mpallansch

mpallansch: Created another PR to address

PR #1051, opened 2024-01-29
[dev-6597] multidashboards

joshlacey:


adamdoe (changes requested):
This is looking great so far! I'm seeing a few issues.

The first issue is that the initial tile/widget icons are now blue. I think they were originally designed to be black until the user chose one of the tiles. 

The second issue I'm seeing is that visualization disappears while editing it within the dashboard section now. For example, while building a bar chart with the example data I'm not able to see the bar chart while editing. I can, however, see it on the multi dashboard screen.

Nice work, I'll dig further after these revisions!

adamdoe:
@joshlacey when I set up an individual visualization inside of a dashboard it doesn't show. For example, setting axis and color values on an individual visualization. I can see the output on the dashboard preview screen but not on the editing screen.

PR #1058, opened 2024-01-31
Global CSS Vars

joshlacey:
Proposal to set all TP4 vars to the global CSS context.
image

1) We almost never need write a hex code or define scss color variables.
2) These vars can be used inside our existing scss.
3) We will be able to use these vars in regular css files.
4) We can access these variables in Javascript:
image

### DIFF
--- a/packages/core/styles/_global-variables.scss
+++ b/packages/core/styles/_global-variables.scss
@@ -1,0 +1,75 @@
+$colors: (
+  'baseColor': #333,
+  'blue': #005eaa,
+  'lightestGray': #f2f2f2,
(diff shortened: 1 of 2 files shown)


adamdoe:
Looks good to me. Were these pulled from TP or our current colors inside of COVE?

joshlacey:
@adamdoe These were pulled from core/styles/_variables.scss . I feel like ultimately it would be good if TemplatePackage exported these variables. We could add these in the interim and them maybe suggest to them that they add these.

adamdoe (approved):

PR #1062, opened 2024-02-01
DEV 6811

mpallansch:
Fixes issues from testing:

1. If you remove the Latitude/Longitude configuration after selecting it, you get an unrecoverable error
2. Can we have the cities we support by default be placed, even if others are placed using lat/long?  Example, "New York City" in the data with no lat/long
3. If you switch to pin, then switch back, the dots disappear because the "geoCodeCircleSize" or some setting is now gone
4. Need to expose the "Geocode Circle Size" setting when lat/long is used.  Could that default to the size of the normal city circles and allow folks to adjust it whether lat/long is used or not?
5. If you filter the map to only show some locations, the custom locations still appear


adamdoe:
@mpallansch - I tested with the world geo code example file. I'm still seeing the circle appear for NYC when I have the location set to United States. Same with having the location set to Alaska.

mpallansch:
> @mpallansch - I tested with the world geo code example file. I'm still seeing the circle appear for NYC when I have the location set to United States. Same with having the location set to Alaska.

Looks like that file has erroneous lat/long data for each point (US, Alaska, and NYC all have NYC lat/long data):

{
      "Country": "New York City",
      "Cases": 300,
      "Category": "Has not historically reported monkeypox",
      "AsOf": "11 Jul 2022 5:00 PM EDT",
      "longitude": "-74.006",
      "latitude": "40.712"
    },
    {
      "Country": "United States of America",
      "Cases": 10,
      "Category": "Has not historically reported monkeypox",
      "AsOf": "11 Jul 2022 5:00 PM EDT",
      "longitude": "-74.006",
      "latitude": "40.712"
    },
    {
      "Country": "Alaska",
      "Cases": 500,
      "Category": "Has not historically reported monkeypox",
      "AsOf": "11 Jul 2022 5:00 PM EDT",
      "longitude": "-74.006",
      "latitude": "40.712"
    }

I don't think this will be an issue for valid use cases.

PR #1066, opened 2024-02-01
[DEV-6472-solr] initial work for solr requests

adamdoe:
### DEV-6472

You should be able to pull example files from the original ticket, save them under a private folder in editor/example/private as `wt=json` or `wt=csv`. You can then mock the request using something like `http://localhost:8080/example/private/wt=csv` Based on if the wt value is found we determine if the solr request is csv or json and get the data.

### DIFF
--- a/packages/editor/src/components/DataImport.jsx
+++ b/packages/editor/src/components/DataImport.jsx
@@ -26,6 +26,20 @@ import '../scss/data-import.scss'
 
 import '@cdc/core/styles/v2/components/data-designer.scss'
 
+const isSolrCsv = dataUrl => {
+  if (dataUrl.includes('wt=csv')) {
+    return true
+  }
+  return false
+}
+
+const isSolrJson = dataUrl => {
+  if (dataUrl?.includes('wt=json')) {
+    return true
+  }
+  return false
+}
+
 export default function DataImport() {
   const { config, setConfig, errors, setErrors, errorMessages, maxFileSize, setGlobalActive, tempConfig, setTempConfig, sharepath } = useContext(ConfigContext)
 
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/DataImport.jsx
```diff
@@ -26,6 +26,20 @@ import '../scss/data-import.scss'
...
+  return false
+}
+
+const isSolrJson = dataUrl => {
+  if (dataUrl?.includes('wt=json')) {
+    return true
+  }
+  return false
+}
+
```
joshlacey, line 42: could you collapse this into one function: 

const getSolrExtension = (dataURL: string) => '.json' | '.csv' | null

packages/editor/src/components/DataImport.jsx
```diff
@@ -81,15 +95,15 @@ export default function DataImport() {
   const loadExternal = async () => {
     let dataURL = ''
     // Is URL valid?
+
     try {
-      dataURL = new URL(externalURL, window.location.origin)
+      dataURL = isSolrCsv(externalURL) || isSolrJson(externalURL) ? externalURL : new URL(externalURL, window.location.origin)
```
joshlacey, line 100: getSolrFileType(externalURL)

packages/editor/src/components/DataImport.jsx
```diff
@@ -100,9 +114,9 @@ export default function DataImport() {
           responseBlob = response.data
 
           // Sometimes the files are coming in as plain text types... Maybe when saved from Macs
-          if (fileExtension === '.csv' && responseBlob.type === 'text/plain') {
+          if ((fileExtension === '.csv' && responseBlob.type === 'text/plain') || isSolrCsv(externalURL)) {
```
joshlacey, line 117: getSolrExtension(exteranlURL) === '.csv'

joshlacey:
not sure what solr is

PR #1073, opened 2024-02-05
Sankey adam

adamdoe:


adamdoe:
@navbajwa97 - this gets the sankey to the point of updating data for both an individual sankey diagram and a sankey inside of the dashboard. 

I included an example dashboard file that looks close to the requirements for this chart type. If you need to make updates to the dashboard I'd recommend doing that inside of dashboard directly and then copy/pasting the config OR re-building the example file using the editor package. 

You might have to re-run the `yarn install` command and then run `lerna run build` to rebuild all packages. Then you can run lerna run --scope @cdc/dashboard start` to see the updates.

PR #1091, opened 2024-02-14
Feature: [dev-7205] stand alone table widget

joshlacey:
- Added stand alone table widget 
- created a updateFieldFactory function which can be used by multiple editors.
- added ColumnsEditor (can be used to refactor other Editors that use Columns configuration)
- added EditorWrapper to Core so that we can keep all editors looking the same ( can be used to refactor current editors)


adamdoe (changes requested):
Nice work with TS and breaking the functions out here as usual. I also like how you created the EditorPanelWrapper file.  

Using the chart example file, here are a few things I'm seeing we should try to address:

- nit: Current editor panels use `Panel.[sectionName]` as their component name, ie. `/packages/chart/components/EditorPanel/components/Panels`
- fix: Review that checkboxes/inputs work as expected within the editor when a table component is created. I didn't do any deep dives here but it might be related to the updateField refactor. Not sure.
- Toggling custom config on and off displays different data in the table (ie. decimals with the default chart file)
- The default table has an extra empty row displaying at the bottom
- Using the exclude columns multi dropdown causes the page to refresh

Thanks again!

joshlacey:
1) I spent a lot of time trying to refactor this and ran into some issues. I can walk you through them, otherwise lets leave this the same.
2) Some of the toggles in the DataTableEditor were actually outside the scope of the DataTable Component so I added a conditional for those.
3) I removed this as the logic is now present in the ColumnsEditor.
4) This is an existing issue in cove that happens anytime you import CSV data. an extra "columns" data point is added.
5) see item 3

PR #1123, opened 2024-02-27
Enhancement: [dev-7287] Allow Data Source URLs which don't end in 'csv' or 'json'

joshlacey:


joshlacey:
https://github.com/CDCgov/cdc-open-viz/blob/ed701e62d4930b96dec603488376e555e964fb31/packages/chart/src/CdcChart.tsx#L175 @mpallansch it seems like this won't work if you use a standalone chart? I'm not sure what the need for the cacheBusting string is but it's not currently used by the chronic cove api.

adamdoe (approved):

PR #1124, opened 2024-02-27
Adds Logic For External Link Indicator

mpallansch:


adamdoe (approved):
Tested on maps and looking good.

PR #1133, opened 2024-03-04
Fix: [dev-7048] Pivot Filter fix after UX comments 

joshlacey:
- request on ticket was made to display no columns when nothing is selected.

- also added update for storybook to include name in the storybook js bundles to give insight into which stories are largest. Dashboard is currently very large so maybe we can figure out how to reduce that

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -47,7 +47,7 @@ const config: StorybookConfig = {
         },
         rollupOptions: {
           output: {
-            chunkFileNames: `storybook-[hash].[format].js`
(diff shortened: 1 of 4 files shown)


adamdoe (changes requested):
I'm not seeing the pivot table update the visualizations or the standalone table component on dashboards. Can you create an example file for dashboards?

joshlacey:
@adamdoe this is the earlier example: https://cdcgov.github.io/cdc-open-viz/?path=/story/components-pages-dashboard--pivot-filter

adamdoe (approved):

PR #1135, opened 2024-03-05
Dev 936

mpallansch:
Allows for deep linking to specific filters in a COVE visualization through the page's query string parameters


packages/core/components/Filters.jsx
```diff
@@ -71,14 +71,30 @@ export const useFilters = props => {
...
     if (visualizationConfig.filterBehavior === 'Apply Button') {
       newFilters[index].queuedActive = value
       setShowApplyButton(true)
     } else {
-      newFilters[index].active = value
+      let newFilter = newFilters[index];
+      newFilter.active = value
+      
+      let queryParams = getQueryParams();
+      if(newFilter.setByQueryParameter && queryParams[newFilter.setByQueryParameter] !== newFilter.active){
```
joshlacey, line 93: can you add this setByQueryParameter to the SharedFilter type?

packages/chart/src/CdcChart.tsx
```diff
@@ -219,6 +219,17 @@ export default function CdcChart({ configUrl, config: configObj, isEditor = fals
       }
     }
     let newConfig = { ...defaults, ...response }
+    if(newConfig.filters){
+      const urlParams = new URLSearchParams(window.location.search);
+      newConfig.filters.forEach((filter, index) => {
+        if(filter.setByQueryParameter && urlParams.get(filter.setByQueryParameter)){
+          if(filter.values && filter.values.includes(urlParams.get(filter.setByQueryParameter))){
+            newConfig.filters[index].active = urlParams.get(filter.setByQueryParameter)
```
joshlacey, line 227: consider pulling urlParam.get(filter.setByQueryParameter) to a variable. This section is a bit dense.

packages/chart/src/components/EditorPanel/EditorPanel.tsx
```diff
@@ -2702,6 +2702,17 @@ const EditorPanel = () => {
                                 />
                               </label>
 
+                              <label>
```
joshlacey, line 2705: Shouldn't this be set by the filter value itself? Like wouldn't the filter have already responded to the query param change?

packages/core/components/Filters.jsx
```diff
@@ -71,14 +71,30 @@ export const useFilters = props => {
 
   const announceChange = text => {}
 
+  const getQueryParams = () => {
+    let queryParams = {};
```
joshlacey, line 75: const

packages/core/components/Filters.jsx
```diff
@@ -71,14 +71,30 @@ export const useFilters = props => {
...
   const changeFilterActive = (index, value) => {
     let newFilters = visualizationConfig.type === 'map' ? [...filteredData] : [...visualizationConfig.filters]
 
     if (visualizationConfig.filterBehavior === 'Apply Button') {
       newFilters[index].queuedActive = value
       setShowApplyButton(true)
     } else {
-      newFilters[index].active = value
+      let newFilter = newFilters[index];
+      newFilter.active = value
```
joshlacey, line 90: const

packages/core/components/Filters.jsx
```diff
@@ -97,12 +113,22 @@ export const useFilters = props => {
   }
 
   const handleApplyButton = newFilters => {
+    let needsQueryUpdate = false;
+    let queryParams = getQueryParams();
```
joshlacey, line 117: const

packages/dashboard/src/helpers/getUpdateConfig.ts
```diff
@@ -24,7 +24,13 @@ export const getUpdateConfig =
           _filter.values = filterValues
           if (filterValues.length > 0) {
             const defaultValues = _filter.pivot ? _filter.values : _filter.values[0]
-            _filter.active = _filter.active || defaultValues
+
+            const urlParams = new URLSearchParams(window.location.search);
+            if(_filter.setByQueryParameter && urlParams.get(_filter.setByQueryParameter) && _filter.values && _filter.values.includes(urlParams.get(_filter.setByQueryParameter))){
```
joshlacey, line 29: Can you add a comment about what this is doing here or maybe abstract it out to a variable as below?
```const descriptionOfCondition = _filter.setByQueryParameter && urlParams.get(_filter.setByQueryParameter) && _filter.values && _filter.values.includes(urlParams.get(_filter.setByQueryParameter))

if(descriptionOFCondition) { // do something

File: packages/map/src/CdcMap.tsx [hunk 8]
```diff
@@ -1449,6 +1448,16 @@ const CdcMap = ({ className, config, navigationHandler: customNavigationHandler,
       filters = generateRuntimeFilters(state, hashFilters, runtimeFilters)
 
       if (filters) {
+        if(JSON.stringify(runtimeFilters) !== JSON.stringify(filters)){
+          const urlParams = new URLSearchParams(window.location.search);
+          filters.forEach((filter, index) => {
+            if(filter.setByQueryParameter && urlParams.get(filter.setByQueryParameter)){
```
joshlacey, line 1454: This logic seems to repeat a lot. Could it be abstracted to a function and added to core?

packages/map/src/components/EditorPanel/components/EditorPanel.tsx
```diff
@@ -1288,6 +1288,9 @@ const EditorPanel = props => {
               })}
             </select>
           </label>
+
```
joshlacey, line 1291: same comment as I had on the chart component. do these components do some filtering on their own somehow?

packages/chart/src/CdcChart.tsx
mpallansch: Moved to shared function and addressed in latest commit

packages/chart/src/components/EditorPanel/EditorPanel.tsx
mpallansch: I'm not sure I understand what you're asking. This section of the code is so that the user can specify in the editor which QS param they want to map to which filter dropdown

packages/core/components/Filters.jsx
mpallansch: Addressed in latest commit

packages/core/components/Filters.jsx
mpallansch: Addressed in latest commit

packages/core/components/Filters.jsx
mpallansch: Addressed in latest commit

packages/dashboard/src/helpers/getUpdateConfig.ts
mpallansch: Moved to a shared function and addressed in latest commit

packages/map/src/CdcMap.tsx
mpallansch: Moved to a shared function and addressed in latest commit

packages/map/src/components/EditorPanel/components/EditorPanel.tsx
mpallansch: See response above.

packages/chart/src/components/EditorPanel/EditorPanel.tsx
joshlacey: I see now that you can set filters directly in the chart editor. that's what I was confused about.

joshlacey:
@mpallansch would it be possible to get a storybook example of this working?

mpallansch:
> @mpallansch would it be possible to get a storybook example of this working?
> 
> I created a mini example of this in editor but I don't see anything changing. I would expect that on load of this page the filter should be set to whatever's in the query parameter. I would also expect that changing the filter option would also update the query parameter in the url
> 
> image

I'm not sure about storybook, but my recent commit added an example config you can use. You will need to specify in the editor  which QS param corresponds to which filter using the new input: 

Screen Shot 2024-03-07 at 2 59 01 PM

joshlacey:
@mpallansch I was able to get the example running I see it changing the query param when the filter is changed, but then when I refresh the page with that query param I'm not seeing the value selected in the dropdown. 

If you could, please try to add this example to storybook it would help me communicate what you're doing to my UX team.

mpallansch:
> @mpallansch I was able to get the example running I see it changing the query param when the filter is changed, but then when I refresh the page with that query param I'm not seeing the value selected in the dropdown.
> 
> If you could, please try to add this example to storybook it would help me communicate what you're doing to my UX team.

You were seeing this issue with the example config that I provided? I'm seeing it work consistently across all browsers for me. @adamdoe are you seeing the same issue as Josh?

joshlacey (approved):

PR #1160, opened 2024-03-11
Feature/dev 5717b

Atash3000:
Fixed bugs on Map city shapes.

### DIFF
--- a/packages/map/src/components/EditorPanel/components/EditorPanel.tsx
+++ b/packages/map/src/components/EditorPanel/components/EditorPanel.tsx
@@ -120,15 +120,8 @@ const EditorPanel = props => {
   const getCityStyleOptions = target => {
     switch (target) {
       case 'value': {
-        let values = ['Circle', 'Square', 'Triangle', 'Diamond', 'Star', 'Pin']
-        const filteredValues = values.filter(val => {
-          let hasMatch = true
-          if (String(state.visual.cityStyle).toLocaleLowerCase() === String(val).toLocaleLowerCase()) {
-            hasMatch = false
-          }
-
-          return hasMatch
-        })
+        const values = ['Circle', 'Square', 'Triangle', 'Diamond', 'Star', 'Pin']
+        const filteredValues = values.filter(val => String(state.visual.cityStyle).toLocaleLowerCase() !== val.toLocaleLowerCase())
 
         return (
           <>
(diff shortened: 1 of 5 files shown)


packages/map/src/components/Legend/components/index.scss
```diff
@@ -189,7 +189,7 @@
           justify-items: center;
           line-break: loose;
           align-items: center;
-          width: auto;
+          width: auto !important;
```
adamdoe, line 192: Avoid using !important keyword in favor for a more specific selector.

packages/map/src/components/Legend/components/Legend.tsx
```diff
@@ -187,30 +187,32 @@ const Legend = () => {
...
-                            {shapes[shape.toLowerCase()]}
-                          </Group>
-                        </svg>
-                        <p>{label}</p>
-                      </div>
-                    )
-                )}
+                <div className={legendClasses.div.join(' ') || ''}>
+                  {state.visual.cityStyleLabel && (
+                    <div>
```
adamdoe, line 192: Is the empty div necessary here?

packages/map/src/components/Legend/components/Legend.tsx
```diff
@@ -187,30 +187,32 @@ const Legend = () => {
...
+                        </Group>
+                      </svg>
+                      <p>{state.visual.cityStyleLabel}</p>
+                    </div>
+                  )}
+
+                  {state.visual.additionalCityStyles.map(
+                    ({ shape, label }) =>
+                      label && (
+                        <div>
```
adamdoe, line 205: Similar, do we need an empty div here?

packages/map/src/components/EditorPanel/components/EditorPanel.tsx
```diff
@@ -119,18 +119,23 @@ const EditorPanel = props => {
...
     switch (target) {
-      case 'column': {
-        return (
-          <option value='' key={'Select Option'}>
-            - Select Option -
-          </option>
-        )
-      }
       case 'value': {
         let values = ['Circle', 'Square', 'Triangle', 'Diamond', 'Star', 'Pin']
```
adamdoe, line 123: Use const here

adamdoe (changes requested):

adamdoe (changes requested):
See feedback

packages/map/src/components/Legend/components/Legend.tsx
Atash3000: We need these because we apply css flex on them to align <p> & svg

packages/map/src/components/Legend/components/index.scss
Atash3000: Fixed.

PR #1170, opened 2024-03-12
Feature/charts-data-series-remaining

petercolella:
I do not have the ticker number, but the description is as follows:

> Currently, the logic for Data Series renaming is that the renaming field, that appears when you expand an added Data Series, shows up depending on the chart you are building when you add the Data Series.
> 
> - For example, when buildng a bar chart, the "Series Name" field appears when you toggle the Data Series open.
> - But if you then switch to scatter plot, add a second Data Series, and expand that Data Series, there is no "Series Name" field to edit.
> - In this ticket, please change the functionality so that the "Series Name" field is shown when you expand the Data Series to see its sub-options.
> - Note: Box Plot does not allow you to expand the Data Series to see sub-options, and that is the expected behavior. Box plot does not show a legend and the hovers are set.

Screenshot 2024-03-12 164651
Screenshot 2024-03-12 164743
Screenshot 2024-03-12 170033
Screenshot 2024-03-12 170300

### DIFF
--- a/packages/chart/src/components/EditorPanel/components/Panels/Panel.Series.tsx
+++ b/packages/chart/src/components/EditorPanel/components/Panels/Panel.Series.tsx
@@ -427,7 +427,7 @@ const SeriesDropdownConfidenceInterval = props => {
 const SeriesInputName = props => {
   const { series, index: i } = props
   const { config, updateConfig } = useContext(ConfigContext)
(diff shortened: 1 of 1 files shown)


petercolella:
@adamdoe

adamdoe (approved):
Looks good, thanks!

PR #1172, opened 2024-03-12
[DEV-7448] Add info text for Map Layers

Atash3000:


packages/map/src/components/EditorPanel/components/EditorPanel.tsx
```diff
@@ -3120,6 +3120,7 @@ const EditorPanel = props => {
                 <button className={'btn full-width'} onClick={handleAddLayer}>
                   Add Map Layer
                 </button>
+                <p>Context should be added to your visualization or associated page to describe the significance of layes that are added to maps.</p>
```
adamdoe, line 3123: layers*

adamdoe (changes requested):

packages/map/src/components/EditorPanel/components/EditorPanel.tsx
adamdoe: @Atash3000

packages/map/src/scss/editor-panel.scss
```diff
@@ -280,6 +280,9 @@
       span.h5 {
         font-size: 0.8em;
       }
+      & p {
```
adamdoe, line 283: I think this should use a more focused selector. I'm not sure we want all paragraph items inside of accordion panels to have a margin top on them.

adamdoe (approved):

PR #1173, opened 2024-03-13
[DEV-7420] Add default Data table Header to Charts

Atash3000:


packages/core/components/EditorPanel/DataTableEditor.tsx
```diff
@@ -75,7 +75,7 @@ const DataTable: React.FC<DataTableProps> = ({ config, updateField, isDashboard,
           }
         />
       )}
-      <TextField value={config.table.indexLabel} section='table' fieldName='indexLabel' label='Index Column Header' updateField={updateField} />
+      <TextField value={config.table.indexLabel ? config.table.indexLabel : config.xAxis?.dataKey} section='table' fieldName='indexLabel' label='Index Column Header ddddd' updateField={updateField} />
```
adamdoe, line 78: @Atash3000 please remove the ddddd part here. I'm also wondering if we should use the data key at all. Most data keys that I've seen aren't the most readable and I'm not sure we want to output that to the user. If I remember correctly, the ticket shared that its okay to output `<span aria-hidden>No title</span>` as a backwards compatible option.

adamdoe (changes requested):

packages/core/components/EditorPanel/DataTableEditor.tsx
Atash3000: If data key is not readable user can override it using Index header input under Data Table section.

packages/core/components/EditorPanel/DataTableEditor.tsx
```diff
@@ -75,7 +75,7 @@ const DataTable: React.FC<DataTableProps> = ({ config, updateField, isDashboard,
           }
         />
       )}
-      <TextField value={config.table.indexLabel} section='table' fieldName='indexLabel' label='Index Column Header' updateField={updateField} />
+      <TextField value={config.table.indexLabel ? config.table.indexLabel : config.xAxis?.dataKey} section='table' fieldName='indexLabel' label='Index Column Header' updateField={updateField} />
```
adamdoe, line 78: I think what we need here is the empty string still. Then, in the table header output we can check for the empty string and if it is empty we can output the content that's mentioned in the ticket `<th><span sr-only>Not Applicable</span></th>`

adamdoe (changes requested):

adamdoe:
@Atash3000 this PR has feedback on it.

PR #1180, opened 2024-03-14
[DEV-7440] Fix tab index

Atash3000:
Remove screen reader from scatter plot circles


adamdoe:
@Atash3000 I think this is already resolved.

PR #1187, opened 2024-03-18
[DEV-7434] change css for Skip Container link

Atash3000:
Hidden tab stop between skip over chart container and filter buttons. Content hidden from sighted users needs to be hidden from all users, made not keyboard focusable.


adamdoe:
Discussed trying a different approach, changing positioning isn't a solution for screen reader only text.

PR #1188, opened 2024-03-18
Feature/dev 5717c

Atash3000:
Add logic to shapes to handle latitude and longitude of the city

### DIFF
--- a/packages/map/src/components/CityList.jsx
+++ b/packages/map/src/components/CityList.jsx
@@ -154,6 +154,16 @@ const CityList = ({ data, state, geoClickHandler, applyTooltipsToGeo, displayGeo
           </g>
         )
       }
+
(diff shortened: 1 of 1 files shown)


packages/map/src/components/CityList.jsx
```diff
@@ -154,6 +154,16 @@ const CityList = ({ data, state, geoClickHandler, applyTooltipsToGeo, displayGeo
           </g>
         )
       }
+
+      if (geoData?.[state.columns.longitude.name] && geoData?.[state.columns.latitude.name]) {
+        let coords = [Number(geoData?.[state.columns.longitude.name]), Number(geoData?.[state.columns.latitude.name])]
+        let translate = `translate(${projection(coords)})`
```
adamdoe, line 160: Use const here. Is coords used anywhere?

adamdoe (changes requested):

PR #1195, opened 2024-03-20
Feature/dev 7418

Atash3000:
Updated initial state for Download Link in Data Table


adamdoe:
@Atash3000 this needs additional work. It's a good start to default everything on, but what if users decide to turn the option off in the editor panel? We'll need to hide those options as well, and use the coveUpdateWorker in core to default previous configs download options to true.

adamdoe (changes requested):
See comment

PR #1239, opened 2024-04-16
fixed storybook build

joshlacey:
Storybook was not building the editor panel Accordion for the standalone table. After these changes it's fixed. I've already rebuilt storybook with this change so that we can have this for the demo tomorrow. No urgency in getting this out.

1) In storybook go to Pages/Dashboard/Pivot Filter.
2) in the config panel below toggle the isEditor to true
3) in dashboard click on the tools icon to bring up the table editor.

Expected: you should see the Columns and Data Table accordions on the left.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     ]
   },
   "scripts": {
+    "clean": "find . -name 'node_modules' -o -name 'dist' -type d -prune -exec rm -rf '{}' +",
     "publish": "npx lerna publish",
     "storybook": "storybook dev -p 6006",
     "storybook-docs": "storybook dev --docs",
--- a/packages/core/components/DataTable/DataTableStandAlone.tsx
+++ b/packages/core/components/DataTable/DataTableStandAlone.tsx
@@ -1,14 +1,24 @@
 import { ViewPort } from '../../types/ViewPort'
 import { Visualization } from '../../types/Visualization'
+import EditorWrapper from '../EditorWrapper/EditorWrapper'
 import DataTable from './DataTable'
+import DataTableEditorPanel from './components/DataTableEditorPanel'
 
 type StandAloneProps = {
   visualizationKey: string
   config: Visualization
   viewport?: ViewPort
+  isEditor?: boolean
+  updateConfig?: Function
 }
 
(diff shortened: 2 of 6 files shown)


adamdoe (changes requested):
I'm not sure what the benefit of the clean command here is. Other than that I think this PR looks good-to-go. I think the `lerna run clean` command should provide what your looking for at the root.

joshlacey:
> I'm not sure what the benefit of the clean command here is. Other than that I think this PR looks good-to-go. I think the `lerna run clean` command should provide what your looking for at the root.

this is the output I get when I run `lerna run clean`

$ lerna run clean
lerna notice cli v6.4.1
lerna info versioning independent
lerna WARN EWORKSPACES Workspaces exist in the root package.json, but Lerna is not configured to use them.
lerna WARN EWORKSPACES To fix this and have Lerna use workspaces to resolve packages, set `useWorkspaces: true` in lerna.json.
lerna WARN EPACKAGES No packages defined in lerna.json. Defaulting to packages in packages/*
lerna success run No packages found with the lifecycle script 'clean'


I need some way of removing the node_modules and dist folders recursively. sometimes I'm running into caching issues when debugging and this is the only way to reset everything. Any thoughts?

mpallansch (approved):

adamdoe:
@joshlacey - I think the answer is in that error message. I think your solution works well if you want to keep it in, otherwise you can temporarily add in the setting `useWorkspaces: true` to the lerna.json file. It gets removed during releases to accomplish versioning/publish tasks where that text needs to be removed for some reason.

adamdoe (approved):

joshlacey:
> @joshlacey - I think the answer is in that error message. I think your solution works well if you want to keep it in, otherwise you can temporarily add in the setting `useWorkspaces: true` to the lerna.json file. It gets removed during releases to accomplish versioning/publish tasks where that text needs to be removed for some reason.

after making that change I'm now getting this: 

$ lerna run clean
lerna notice cli v6.4.1
lerna info versioning independent
lerna success run No packages found with the lifecycle script 'clean'

adamdoe:
@joshlacey use `lerna clean`

PR #1249, opened 2024-04-23
Enhancement: [DEV-7807] standalone table additional enhancements

joshlacey:
Testing Scenario setup: 
- User is in COVE Editor
- Select Dashboard Visualization
- Select any data source
- drag and drop a standalone data table to the visualization panel and configure the data source
- click to tool icon to bring up the table editor panel. 

** these controls were added to Chart as well so additional regression testing may be required there

testing instructions:
1) Set Column order:

- open the column accordion on the left and select "add configuration"
- select a column to configure and select the order you'd like to see that column

Expected: you should see the column shift position to the expected order (example 3 means that the column should be the 3rd column in the data table)

2) Hide Columns
- open the column accordion on the left and select "add configuration"
- select a column to configure and toggle the "show in data table" check box off

Expected: you should see the column removed from the data table.

- open the data table accordion.

Expected: under "exclude column" you should see the column name you selected above.

3) Data filters added to standalone table

Expected:  open filter accordion and configure a filter as you would on any other visualization

4) Remove unused Data Table controls from Data Table editor
** this test does not apply to Charts
Expected: you should see the following controls: 
- Data Table Title
- Screen Reader Description
- Limit Table Height
- Exclude Columns
- Collapsible
- Expanded By Default
- Table Cell Min Width

5) Make Data Tables non Collapsible
- Open Data Table accordion
- uncheck the collapsible checkbox.

expected: you should see the accordion bar at the top of the table removed and you should see the option to check "expanded by default" to be removed from the Data Table configuration panel.

6) Group Rows
- Open Data Table Accordion.
- Select a Row to group by.

Expected: you should see rows grouped by the selected columns


packages/core/components/DataTable/DataTableStandAlone.tsx
```diff
@@ -1,15 +1,52 @@
...
+  setConfig: (Visualization) => void
 }
 
-const DataTableStandAlone: React.FC<StandAloneProps> = ({ visualizationKey, config, viewport }) => {
-  return <DataTable expandDataTable={true} config={config} rawData={config.data} runtimeData={config.formattedData} tabbingId={visualizationKey} tableTitle={config.table.label} viewport={viewport || 'lg'} />
+// filterData is copied from ./packages/chart/src/helpers/filterData.ts
+// consider moving this to a shared location
+const filterData = (filters, data) => {
+  if (!filters) return data
+  let filteredData: any[] = []
```
mpallansch, line 18: let could be const here

mpallansch (commented):
Everything looked good in testing except for the filter. I tried using the default config file referenced in index.html (/examples/full-dashboard.json) and added a standalone table, added filter, selected year for the column. The dropdown does show, but it doesn't filter the table to rows where the year is the one selected. This is the case for all columns, and for apply/change behavior.

joshlacey:
> Everything looked good in testing except for the filter. I tried using the default config file referenced in index.html (/examples/full-dashboard.json) and added a standalone table, added filter, selected year for the column. The dropdown does show, but it doesn't filter the table to rows where the year is the one selected. This is the case for all columns, and for apply/change behavior.

thanks for catching this @mpallansch I believe I've resolved the issue in a commit I just pushed up.

mpallansch (approved):

PR #1250, opened 2024-04-24
refactor: [dev-7142] world map to typescript

joshlacey:
this is preliminary work for dev-7142 so that I can make changes that we can track since the World map file is moving.


adamdoe:
@joshlacey the map package fails to build. Remember to test all the way through editor.

joshlacey:
> @joshlacey the map package fails to build. Remember to test all the way through editor.

I tested the editor and everything was working. I did see when I ran the build again that the import for the topo.json file wasn't correct, IDK why it didn't throw an error sooner than running the build.

just pushed the change up

adamdoe (approved):

PR #1272, opened 2024-05-13
Feature: [dev-7696] dynamic footnotes

joshlacey:


packages/core/components/EditorPanel/FootnotesEditor.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const getOptions = (opts: string[]) => {
+    return [['', '--Select--']].concat(opts.map(key => [key, key]))
+  }
+
+  const datasets = config.datasets || {}
+
+  const dataColumns = config.dataKey ? getOptions(Object.keys(datasets[config.dataKey]?.data?.[0] || {})) : []
+  const dataSetOptions = getOptions(Object.keys(datasets))
+  return (
+    <>
```
adamdoe, line 39: Implement `<Layout.Sidebar />` here so we have similar styles/visuals throughout the app.

packages/core/components/Footnotes/footnotes.css
```diff
@@ -0,0 +1,5 @@
+:is(ul) {
```
adamdoe, line 1: Should this have an outer wrapper on it? Maybe cdc-open-viz-module or a footnotes class?

packages/dashboard/src/components/Row.tsx
```diff
@@ -149,31 +150,52 @@ const RowMenu: React.FC<RowMenuProps> = ({ rowIdx }) => {
...
+        </button>
+        <div className='column-container'>
           {row.columns
             .filter(column => column.width)
             .map((column, colIdx) => (
               <Column data={column} key={`row-${uuid}-col-${colIdx}`} rowIdx={rowIdx} colIdx={colIdx} />
             ))}
-        </>
+        </div>
+        <button className='btn btn-primary footnotes' onClick={configureFootnotes}>
```
adamdoe, line 194: Should this have a state to show that footnotes have been configured. Maybe Edit Footnotes or a solid color to show that they've been setup?

packages/core/components/EditorPanel/FootnotesEditor.tsx
joshlacey: It looks like <Layout.Sidebar> is duplicating some of what EditorWrapper was doing. I've used <EditorWrapper> Around this. I think we need to do a separate refactor of EditorWrapper for consistency, and I don't want to put that in the scope of this PR.

PR #1273, opened 2024-05-13
[DEV-8167] Hides Elements from Image Download

mpallansch:


packages/core/components/MediaControls.jsx
```diff
@@ -59,8 +59,9 @@ const generateMedia = (state, type, elementToCapture) => {
 
   switch (type) {
     case 'image':
-      html2canvas(baseSvg).then(canvas => {
-        saveImageAs(canvas.toDataURL(), filename + '.png')
+      html2canvas(baseSvg, {ignoreElements: el => el.className?.indexOf && el.className.search(/download-buttons|download-links|data-table-container/) !== -1}).then(canvas => {
+        document.body.appendChild(canvas);
+        //saveImageAs(canvas.toDataURL(), filename + '.png')
```
adamdoe, line 64: @mpallansch did you mean to comment out the save function here? I'm seeing the save image button removed, which I think is intentional, but I think the save function should still be in place for the downloads folder.

adamdoe (changes requested):
@mpallansch - did you mean to comment out the save function here? I'm seeing the save image button removed, which I think is intentional, but I think the save function should still be in place for the downloads folder.

PR #1274, opened 2024-05-13
Fixes Superscript in Header

mpallansch:


adamdoe:
Tested and working as intended with the rule removed.

PR #1275, opened 2024-05-13
Dev 7840

mpallansch:


adamdoe:
@mpallansch - It looks like @joshlacey has another PR open here that fixes this.
https://github.com/CDCgov/cdc-open-viz/pull/1277

PR #1276, opened 2024-05-14
[shared-filter-dashboards] bar chart color fix

adamdoe:


mpallansch (approved):
It seems the `if` check above was meant to prevent getting into this block in the first place without sharedFilters. It might make sense to change it to this instead:

`if (dashboardConfig && dashboardConfig.dashboard.sharedFilters && dashboardConfig.dashboard.sharedFilters?.length !== 0) {`

Either way it tests well for me and it looks good to go out.

PR #1298, opened 2024-05-22
enhancement: [Dev-8008] trendline bar data driven switch

Matt-DiPaolo:
In storybook under Chart, there are "Line Chart Two Points New Chart" and "Line Chart Two Points Regression Test"
 
Both have the same dataset but the "Regression Test" does not have the config property of `allowLineToBarGraph` This is to simulate how current Line Charts will respond to the update. You will see that the line chart will persist even when there are only 2 data points. You will have to go into the editor, under the General Panel and select the "Convert to Bar Graph" checkbox for the conversion to take place.   

Testing: 
1. Select the "Line Chart Two Points New Chart" under Chart. 
2. Switching the filter from Home to Work and back again should show the Line turning to a Bar chart when there are 2 points and back to a line when there are more than 2 points. 
3. Under Storybook controls, turn isEditor to true
4. In the General Panel the Chart type should be set to Line
5. By toggling the "CONVERT TO BAR GRAPH" checkbox, the graph with 2 points should toggle between a line and bar graph.
6. All other functionality remains the same.
7. Switch Chart Type to Bar
8. Test changing the Subtype, orientation, bar style, and any other Bar properties. 
9. Then change Chart Type back to Line and all of the Bar properties should be reset
10. The Bar Chart that is shown when Line Chart is selected should be a Subtype of Regular, Orientation of Vertical, and the Bar Style of Flat.

The previous steps can be repeated on the "Line Chart Two Points Regression Test." The only difference will be, on initial load there will be no bar charts with the 2 data points. Once the isEditor is set to true and the "CONVERT TO BAR GRAPH" checkbox is checked, then everything should be identical to the "Line Chart Two Points New Chart" steps. 

I also made a change in LinearChart that will be systemic for all charts. `if (orientation === 'vertical' && max - min < 3) return formatNumber(tick, 'left', shouldAbbreviate, false, false, '1')`

The above change will make it so if the difference of max and min value of the yAxis is less than 3, then the yAxis points will round to the .5 . I noticed in the example that the max yAxis was 12 and the min was 10. The left axis had 5 ticks and it showed as [10, 10, 11, 11, 12]. After that change it now show as [10, 10.5, 11, 11.5, 12]


packages/chart/src/CdcChart.tsx
```diff
@@ -481,13 +486,13 @@ export default function CdcChart({ configUrl, config: configObj, isEditor = fals
       })
     }
 
-    if ((newConfig.visualizationType === 'Bar' && newConfig.orientation === 'horizontal') || ['Deviation Bar', 'Paired Bar', 'Forest Plot'].includes(newConfig.visualizationType)) {
+    if (((newConfig.visualizationType === 'Bar' || config.displayLineGraphAs === 'Bar') && newConfig.orientation === 'horizontal') || ['Deviation Bar', 'Paired Bar', 'Forest Plot'].includes(newConfig.visualizationType)) {
```
joshlacey, line 489: could displayLineGraphAsBar be a boolean?

packages/chart/src/types/ChartConfig.ts
```diff
@@ -96,6 +96,7 @@ type AllChartsConfig = {
   dataFormat: DataFormat
   dataKey: string
   description: string
+  displayLineGraphAs: 'Line' | 'Bar'
```
joshlacey, line 99: make this a boolean  and change the variable to convertLineToBarGraph. then put a comment next to this type that explains that this is only for when there's less than 3 values.

packages/chart/src/types/ChartConfig.ts
```diff
@@ -188,8 +189,10 @@ export type ForestPlotConfig = {
 } & AllChartsConfig
 
 export type LineChartConfig = {
-  visualizationType: 'Line'
+  convertLineToBarGraph: boolean
+  displayLineGraphAs: 'Line'
```
joshlacey, line 193: why do you need both of these?

packages/dashboard/src/CdcDashboardComponent.tsx
```diff
@@ -516,6 +516,7 @@ export default function CdcDashboard({ initialState, isEditor = false, isDebug =
 
         switch (visualizationConfig.type) {
           case 'chart':
+            visualizationConfig.visualizationType = visualizationConfig.visualizationType === 'Line' && visualizationConfig.formattedData?.length <= 2 ? 'Bar' : visualizationConfig.visualizationType
```
joshlacey, line 519: I'm not sure why this is necessary. I would think the Chart component could handle this logic itself?

packages/dashboard/src/components/VisualizationsPanel.tsx
```diff
@@ -17,6 +17,8 @@ const addVisualization = (type, subType) => {
 
   switch (type) {
     case 'chart':
+      newVisualizationConfig.convertLineToBarGraph = undefined
+      newVisualizationConfig.displayLineGraphAs = 'Line'
```
joshlacey, line 21: handle this logic in the chart component itself.

packages/chart/src/components/LinearChart.jsx
```diff
@@ -667,19 +672,25 @@ const LinearChart = props => {
...
-              <LineChart xScale={xScale} yScale={yScale} getXAxisData={getXAxisData} getYAxisData={getYAxisData} xMax={xMax} yMax={yMax} seriesStyle={config.series} />
-            </>
-          )}
+          {(visualizationType !== 'Bar' || displayLineGraphAs !== 'Bar') &&
+            visualizationType !== 'Paired Bar' &&
+            visualizationType !== 'Box Plot' &&
+            visualizationType !== 'Area Chart' &&
+            visualizationType !== 'Scatter Plot' &&
+            visualizationType !== 'Deviation Bar' &&
+            visualizationType !== 'Forecasting' && (
```
joshlacey, line 681: I know you didn't write all this but could you try refactoring this to 

!['Paired Bar', 'Box Plot', 'Area Chart', ....].includes(visualizationType)

packages/chart/src/components/EditorPanel/EditorPanel.tsx
```diff
@@ -379,14 +379,15 @@ const EditorPanel = () => {
...
-          type: config.visualizationType === 'Combo' ? 'Bar' : config.visualizationType ? config.visualizationType : 'Bar',
+          type: config.visualizationType === 'Combo' || displayLineGraphAs === 'Bar' ? 'Bar' : config.visualizationType ? config.visualizationType : 'Bar',
           axis: 'Left'
         }
       })
     }
 
     updateConfig({
       ...config,
+      displayLineGraphAs: displayLineGraphAs,
```
joshlacey, line 390: you can just do 

      ...config,
      displayLineGraphAs,
      
   FYI

packages/chart/src/components/BarChart/components/BarChart.Vertical.tsx
```diff
@@ -135,7 +135,7 @@ export const BarChartVertical = () => {
                      * color the bar that is using the filter with barColor and
                      * color the filteredOut (typically gray) bars with the filteredOutColor
                      */
-                    if (dashboardConfig && dashboardConfig.dashboard.sharedFilters?.length !== 0) {
+                    if (dashboardConfig && dashboardConfig.dashboard.sharedFilters?.length !== 0 && dashboardConfig.dashboard.sharedFilters !== undefined) {
```
joshlacey, line 138: I think that question marker after sharedFilters should catch the undefined?

packages/chart/src/CdcChart.tsx
```diff
@@ -1139,9 +1144,17 @@ export default function CdcChart({ configUrl, config: configObj, isEditor = fals
               <SkipTo skipId={handleChartTabbing} skipMessage='Skip Over Chart Container' />
               {/* Visualization */}
               {config?.introText && config.visualizationType !== 'Spark Line' && <section className='introText'>{parse(config.introText)}</section>}
-              <div className={getChartWrapperClasses().join(' ')}>
-                {/* All charts except sparkline */}
-                {config.visualizationType !== 'Spark Line' && chartComponents[config.visualizationType]}
+
+              <div style={{ marginBottom: computeMarginBottom(config, legend, currentViewport) }} className={getChartWrapperClasses().join(' ')}>
+                {/* All charts except line and sparkline */}
+                {config.visualizationType !== 'Spark Line' && config.visualizationType !== 'Line' && chartComponents[config.visualizationType]}
```
joshlacey, line 1150: can you refactor chartComponents to be ChartComponents so that at least it's obvious it's dealing with components?

packages/chart/src/CdcChart.tsx
```diff
@@ -1139,9 +1144,17 @@ export default function CdcChart({ configUrl, config: configObj, isEditor = fals
...
+
+                {/* Line Chart */}
+                {config.displayLineGraphAs === 'Line' && config.visualizationType === 'Line' && chartComponents['Line']}
+
+                {/* Line Chart to Bar Chart */}
+                {config.displayLineGraphAs === 'Bar' && config.visualizationType === 'Line' && ((config.visualizationSubType = 'Regular'), (config.barStyle = 'flat'), (config.isLollipopChart = false), chartComponents['Bar'])}
```
joshlacey, line 1156: Do not set config directly

packages/chart/src/components/LinearChart.jsx
```diff
@@ -40,6 +40,7 @@ const LinearChart = props => {
   const { transformedData: data, tableData, dimensions, config, parseDate, formatDate, currentViewport, formatNumber, handleChartAriaLabels, updateConfig, handleLineType, getTextWidth, brushConfig } = useContext(ConfigContext)
   // todo: start destructuring this file for conciseness
   const { visualizationType, visualizationSubType, orientation, xAxis, yAxis, runtime, debugSvg } = config
+  const displayLineGraphAs = config.convertLineToBarGraph ? config.displayLineGraphAs : 'Line'
```
joshlacey, line 43: ?

packages/chart/src/components/LinearChart.jsx
```diff
@@ -98,10 +99,14 @@ const LinearChart = props => {
       //when logarithmic scale applied change value of first tick
       tick = 0
     }
+    // const leftValues = data.map(datum => Number(datum[config.series[0].dataKey]))
```
joshlacey, line 102: remove comments

packages/chart/src/types/ChartConfig.ts
joshlacey: remove displayLineGraphAs from this whole PR and replace with convertLineToBarGraph

joshlacey (changes requested):

packages/chart/src/types/ChartConfig.ts
Matt-DiPaolo: The convertLineToBarGraph is the checkbox value that is giving permission to change the line to a bar. Should I rename it to allowLineToBeConvertedToBar or something like that? Or do you believe I can get around not using it at all?

packages/chart/src/CdcChart.tsx
Matt-DiPaolo: I've tried to update the config in  updateConfig() and loadConfig(). But when I remove these direct config settings, it causes the chart to not switch from line to bar or bar to line.

adamdoe (changes requested):
@Matt-DiPaolo this looks great! Can you update the mock example imports to be case sensitive? With the casing off on the file names, I believe Mac users will fail to build Storybook. After that, we should be able to merge.

adamdoe (approved):

PR #1302, opened 2024-05-23
Dev 8254

mpallansch:


adamdoe (changes requested):
I'm still seeing the visualizations in the config after removing a row. Here's how I tested:

- lerna run build
- lerna run --scope @cdc/editor start
- Choose dashboard
- Create two rows and add placeholder visualizations to the rows. I don't believe I configured them at this step.
- Remove the last row
- Visualizations are showing in the config.

mpallansch:
> I'm still seeing the visualizations in the config after removing a row. Here's how I tested:
> 
> * lerna run build
> * lerna run --scope @cdc/editor start
> * Choose dashboard
> * Create two rows and add placeholder visualizations to the rows. I don't believe I configured them at this step.
> * Remove the last row
> * Visualizations are showing in the config.

Are the visualizations that are still showing in the config assigned to the first row? It should only wipe the visualizations for the row that is being deleted. I'm seeing it working correctly on my end

PR #1303, opened 2024-05-23
Dev 8102

mpallansch:


packages/chart/src/scss/main.scss
```diff
@@ -82,6 +82,12 @@
     margin-top: 3em;
   }
 
+  .type-pie {
```
adamdoe, line 85: This is impacting all svgs on the page under type-pie, ie. the data table svg plus/minus icon.

packages/chart/src/CdcChart.tsx
```diff
@@ -1101,6 +1101,11 @@ export default function CdcChart({ configUrl, config: configObj, isEditor = fals
   // Prevent render if loading
   let body = <Loading />
```
adamdoe, line 1103: nit: classes in this area could follow the ['className', 'className-2'].join(' ') pattern. There might also be a new layout container to use here, but that could be out of scope for this ticket.

adamdoe (changes requested):

adamdoe (approved):

PR #1312, opened 2024-05-29
[DEV-7540] Remove responsive fonts from legend

Atash3000:


packages/map/src/components/Legend/components/Legend.tsx
```diff
@@ -181,11 +179,7 @@ const Legend = forwardRef((props, ref) => {
         <aside id={skipId || 'legend'} className={legendClasses.aside.join(' ') || ''} role='region' aria-label='Legend' tabIndex={0} ref={ref}>
           <section className={legendClasses.section.join(' ') || ''} aria-label='Map Legend'>
             {legend.title && <h3 className={legendClasses.title.join(' ') || ''}>{parse(legend.title)}</h3>}
-            {legend.dynamicDescription === false && legend.description && (
```
adamdoe, line 184: Why are legend descriptions removed?

adamdoe (changes requested):

packages/map/src/components/Legend/components/Legend.tsx
Atash3000: It is not removed but className removed. since prettier formatter made it one line code it looks like removed. but it is there

PR #1317, opened 2024-05-30
Feature/dev 4215

Atash3000:


packages/chart/src/components/ZoomBrush.tsx
```diff
@@ -27,11 +27,12 @@ const ZoomBrush: FC<Props> = props => {
     endValue: ''
   })
 
+  console.log(textProps, 'text')
```
adamdoe, line 30: Nit: remove console logs

PR #1321, opened 2024-06-03
Keeps Thicker Outline on Selected State When Hover

mpallansch:


adamdoe (changes requested):
I think I might be seeing the issue still on this map:

`<div class="react-container react-container--maps" data-config="https://www.cdc.gov/wcms/4.0/cdc-wp/data-presentation/examples/US-County-Level-Map.json"></div>`

On the initial zoom, it feels hard to distinguish the counties until the user hovers over them. Do we need to address the initial zoom state as well?

PR #1325, opened 2024-06-03
[dev-7840][dev-6574] Api filter Fixes

joshlacey:


joshlacey:
@adamdoe can you check this PR against the bug on DEV-6574? I believe I have the fix. I had to remove the custom js and also delete the hidden filter from the config.

packages/dashboard/examples/ed-visits-county-file.json
```diff
@@ -1,712 +1,618 @@
...
-            {
-                "width": 12,
-                "widget": "chart1701791122773"
-            },
-            {
-                "equalHeight": false
-            },
-            {},
-            {}
+  "dashboard": {
```
joshlacey, line 2: @mpallansch do you have some eslint config that's different from the code base? It reformatted this after I saved.

packages/dashboard/examples/ed-visits-county-file.json
```diff
@@ -1,712 +1,618 @@
...
-                "values": [
-                    "All"
-                ],
-                "active": "All",
-                "tier": 2,
-                "parents": [
-                    "Dashboard Filter 3"
-                ]
-            },
-            {
```
joshlacey, line 111: @mpallansch just making sure you removed this Dashboard Filter 3 from the page you built for Cailin on Test

packages/dashboard/examples/ed-visits-county-file.json
mpallansch: Not that I'm aware of, I don't think I've had this issue before

packages/dashboard/examples/ed-visits-county-file.json
mpallansch: Yes, this filter was removed on the test environment

mpallansch (commented):
I'm seeing an issue when testing the respiratory example. Steps to reproduce:

1. Select Colorado for first dropdown
2. Select Boulder for second dropdown
3. Change first dropdown to another state, like Alaska
4. Chart is blank

I think the chart holds onto the Boulder selection. When the logic is refreshing the dropdown values, it should also reset the active filter if its not in the list anymore

packages/dashboard/examples/ed-visits-county-file.json
```diff
@@ -1,712 +1,618 @@
...
+      {
+        "key": "County",
+        "type": "datafilter",
+        "columnName": "county",
+        "showDropdown": true,
+        "usedBy": ["chart1701791122773", "filtered-text1701791173259", "filtered-text1701791674571", "filtered-text1701791683990", "filter-dropdowns1701792379133", "filtered-text1701888972205", "filtered-text1701888980655"],
+        "values": ["All"],
+        "active": "All",
+        "tier": 2,
+        "parents": ["State"]
```
joshlacey, line 92: @mpallansch I just caught that this also needed to be changed in the config to State, it was still set to Dashboard 3. 

I made a couple other changes as well.

mpallansch (approved):

PR #1334, opened 2024-06-07
[DEV-7795] Annotations

adamdoe:


mpallansch (commented):
Some things I've seen during testing. I'm not sure how the features are supposed to work, so these may not be real issues, feel free to ignore:

1. Opacity slider doesn't seem to change anything for me
2. Connection location doesn't seem to change anything
3. Changing the connection type and connection location and then using snap to location can create gaps between the points 
4. Tooltips seem to interfere with drag and drop functionality

PR #1339, opened 2024-06-07
This is freeze header for module for maps /editor

ernestweems:
@adamdoe  please see commit

### DIFF
--- a/packages/core/components/DataTable/DataTable.tsx
+++ b/packages/core/components/DataTable/DataTable.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState, useMemo } from 'react'
+import { useEffect, useState, useMemo, useRef } from 'react'
 import { timeParse } from 'd3-time-format'
 
 import ErrorBoundary from '@cdc/core/components/ErrorBoundary'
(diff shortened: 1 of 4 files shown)


packages/core/components/DataTable/DataTable.tsx
```diff
@@ -125,6 +125,16 @@ const DataTable = (props: DataTableProps) => {
     maxHeight: config.table.limitHeight && `${config.table.height}px`,
     OverflowY: 'scroll'
   }
+  const [height, getTableHeight] = useState(0)
```
adamdoe, line 128: These three react hooks shouldn't be called conditionally. Use a more descriptive variable here, `height` could refer to a number of different heights throughout the app. Change `getTableHeight` to `setTableHeight` to follow React conventions.

packages/core/components/DataTable/DataTable.tsx
```diff
@@ -125,6 +125,16 @@ const DataTable = (props: DataTableProps) => {
...
+  const tableHgt = useRef(null)
+  useEffect(() => {
+    getTableHeight(tableHgt.current.clientHeight)
+  }, [])
+
+  const fixedHeight = {
```
adamdoe, line 134: The ref is always returning zero for me here, so the table is not displaying for me.

packages/core/components/DataTable/DataTable.tsx
```diff
@@ -125,6 +125,16 @@ const DataTable = (props: DataTableProps) => {
     maxHeight: config.table.limitHeight && `${config.table.height}px`,
     OverflowY: 'scroll'
   }
+  const [height, getTableHeight] = useState(0)
+  const tableHgt = useRef(null)
```
adamdoe, line 129: Best to spell out `tableHeight`. Its a few extra characters but will avoid confusion when other developers work on the app later on.

adamdoe (changes requested):

packages/core/components/DataTable/DataTable.tsx
ernestweems: ok will reach in Teams

PR #1353, opened 2024-06-14
[DEV-7286] 4.24.7 > Inclusive Quantile Legend Opt In

adamdoe:
**Problem**
Equal Number Interval (Opt In) counties were showing as blank counties/states (black on the map) because the max value was calculated using `-1` instead of a decimal value (non inclusive value). Now the scaling accounts for decimal values. There is a storybook page created in this PR. It doesn't seem to respond well and I can't figure out why (file size maybe), so it might be best to test with the example file.

**Changes**
- Used coveUpdateWorker to default the roundToPlace value to zero if its undefined on equal number.
- Updated the domain and breaks values to be inclusive of decimals if equal number opt in is used.
- Sets initial state roundToPlace value to zero for new maps being generated.

**Test Steps**
- Create a county map
- Under the legend panel choose "Equal Number Opt In"
- Under Columns > Data Column > Adjust the round to value to be inclusive for the data.


adamdoe:
@joshlacey @mpallansch

packages/core/helpers/ver/4.24.7.ts
```diff
@@ -0,0 +1,19 @@
+import _ from 'lodash'
+
+const update_4_24_7 = config => {
+  const ver = '4.24.7'
+
+  const newConfig = _.cloneDeep(config)
+
+  // When switching between old version of equal number, and the revised equal number opt in, roundToPlace needs to be set.
```
joshlacey, line 8: @adamdoe could you break the map change out into a function? similar to how the previous ones had been done. I have some changes for 4.24.7 also so there will be conflicts when dashboard filter stuff gets merged. if there are separate functions its a little easier to isolate.

packages/map/src/CdcMap.tsx
```diff
@@ -681,14 +681,28 @@ const CdcMap = ({ className, config, navigationHandler: customNavigationHandler,
         let colors = colorPalettes[state.color]
         let colorRange = colors.slice(0, state.legend.numberOfItems)
 
+        const getDomain = () => {
+          // backwards compatibility
+          if (state?.columns?.primary?.roundToPlace !== undefined && state?.general?.equalNumberOptIn) {
+            return [...new Set(dataSet.map(item => Number(item[state.columns.primary.name]).toFixed(Number(state?.columns?.primary?.roundToPlace))))]
```
joshlacey, line 687: lodash has a _.uniq function which is little cleaner than using Set

joshlacey:
@adamdoe I don't think we should add 85k lines of mock data.

adamdoe:
Thanks @joshlacey, updates are on the last commit.

joshlacey:
@adamdoe do we need the new json config in two places? could one import from the other?

adamdoe:
@joshlacey - I removed the duplicate in favor of the mock data in storybook.

joshlacey (approved):

PR #1372, opened 2024-06-21
feature: [dev-8574] Advanced Editor updates

joshlacey:
We needed a way to more easily configure and edit configs directly in COVE. I've added a upload config button to the first tab of the editor
custom upload

I've also added a new component to the Advanced Editor https://carlosnz.github.io/json-edit-react/ which allows collapse and expand to more granularly edit configs. 
advanced-editor

I've also enhanced the editor to prefilter the config so you're only seeing what will eventually be saved. Before there were temporary data states present which made the config difficult to scroll through.


packages/dashboard/src/store/dashboard.reducer.ts
```diff
@@ -52,6 +53,11 @@ const reducer = (state: DashboardState, action: DashboardActions): DashboardStat
       const [config, filteredData] = getUpdateConfig(state)(...action.payload)
       return { ...state, config, filteredData }
     }
+    case 'APPLY_CONFIG': {
```
adamdoe, line 56: I'm not seeing this action working.

packages/editor/src/components/ChooseTab.tsx
```diff
@@ -309,6 +326,21 @@ export default function ChooseTab() {
           </Tooltip>
         </li>
       </ul>
+      <hr />
```
adamdoe, line 329: This looks good and is working for me. When I used this option I did notice that the Advanced Settings seemed to run really slow. I'm not sure if it was the config I tried uploading. However, compared to the text based format this does seem slower. I think we'll want to weigh pros/cons with speed with text based editor and using this library.

adamdoe (changes requested):
I think the only issue that needs to be resolved here is the apply button. I'm not able to change the config using the new advanced editor at the moment. I think we'll also want to weight the pros/cons of the speed of the advanced editor vs the text based one.

packages/editor/src/components/ChooseTab.tsx
joshlacey: I think maybe because it has to also do a full data pull with this option? with the text editor you're changing things after the data has been updated, which i also thought was no ideal UX because you can also change the data there.

adamdoe (changes requested):
When I paste in an example dashboard config the visualizations property is being stripped and returns an empty dashboard.

joshlacey:
@adamdoe what do you mean paste? the only options should not be to either upload (screenshot 1) or edit (screenshot 2)

adamdoe:
> @adamdoe what do you mean paste? the only options should not be to either upload (screenshot 1) or edit (screenshot 2)

There is an option to edit the config. When I click on the pencil icon for the full config and replace it the previous config is still used.

joshlacey:
> > @adamdoe what do you mean paste? the only options should not be to either upload (screenshot 1) or edit (screenshot 2)
> 
> @joshlacey There is an option to edit the config. When I click on the pencil icon for the full config and replace it the previous config is still used.

did you also hit apply?

adamdoe:
> > > @adamdoe what do you mean paste? the only options should not be to either upload (screenshot 1) or edit (screenshot 2)
> > 
> > 
> > @joshlacey There is an option to edit the config. When I click on the pencil icon for the full config and replace it the previous config is still used.
> 
> did you also hit apply?

Correct here are the issues I'm seeing with steps:

### Issue 1. Loading a config from the starting tile screen
- Rebuild local
- Run the editor
- Upload a working chart config from the examples folder
- Config does not populate if I return to step one
- Config does not load data in screen two
- Config asks to choose a visualization type if I skip to step 3

### Issue 2. Loading a config from the json editor
- Rebuild local
- Run the editor > Choose an example config > Land on step 3
- Scroll to Advanced Editor, on the top level edit the config
- Upload an example working config from chart or map package that matches the initial tile selected
- Hit Apply
- I believe the prior config loads in at this point. When I checked the title properties I wasn't seeing a new title populated and the config was empty.

PR #1374, opened 2024-06-21
bug: [dev-8575] ardi multi viz

joshlacey:
Steps to test are listed in the ticket

### DIFF
--- a/packages/dashboard/src/CdcDashboardComponent.tsx
+++ b/packages/dashboard/src/CdcDashboardComponent.tsx
@@ -94,15 +94,15 @@ export default function CdcDashboard({ initialState, isEditor = false, isDebug =
     return state.config.dashboard.sharedFilters.map((f, i) => i).filter(i => !autoLoadViz.hide.includes(i))
   }, [state.config.dashboard.sharedFilters, state.config.visualizations])
 
-  const setAutoLoadDefaultValue = (sharedFilterIndex: number, filterDropdowns: DropdownOptions, sharedFilters): SharedFilter => {
+  const setAutoLoadDefaultValue = (sharedFilterIndex: number, dropdownOptions: DropdownOptions, sharedFilters): SharedFilter => {
     const sharedFilter = _.cloneDeep(sharedFilters[sharedFilterIndex])
-    if (!autoLoadFilterIndexes.length || !filterDropdowns) return sharedFilter // no autoLoading happening
+    if (!autoLoadFilterIndexes.length || !dropdownOptions) return sharedFilter // no autoLoading happening
     if (!sharedFilter.active && autoLoadFilterIndexes.includes(sharedFilterIndex)) {
       const filterParents = sharedFilters.filter(f => sharedFilter.parents?.includes(f.key))
       const notAllParentFiltersSelected = filterParents.some(p => !(p.active || p.queuedActive))
       if (filterParents && notAllParentFiltersSelected) return sharedFilter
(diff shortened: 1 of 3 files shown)


packages/editor/src/CdcEditor.tsx
```diff
@@ -45,15 +45,15 @@ const CdcEditor: React.FC<WCMSProps> = ({ config: configObj, hostname, container
...
   }
 
   const resizeObserver = new ResizeObserver(entries => {
     const container = entries[0]
-    let { width } = container.contentRect
-    let newViewport = getViewport(width)
-    dispatch({ type: 'EDITOR_SET_VIEWPORT', payload: newViewport })
+    const { width } = container.contentRect
+    const newViewport = getViewport(width)
+    if (state.currentViewport !== newViewport) dispatch({ type: 'EDITOR_SET_VIEWPORT', payload: newViewport })
```
joshlacey, line 56: @mpallansch @adamdoe adding this fix resolved the memory leak. If this won't cause any other issues I think I can clean up this PR and put it out of draft.

adamdoe: I'm not sure how we didn't see that leak until now, this definitely seems necessary. I'm not seeing any issues with the addition.

packages/dashboard/src/CdcDashboardComponent.tsx
```diff
@@ -358,7 +357,8 @@ export default function CdcDashboard({ initialState, isEditor = false, isDebug =
         if (Array.isArray(value)) throw Error(`Cannot set active values on urlfilters. expected: ${JSON.stringify(value)} to be a single value.`)
         newSharedFilters[index].queuedActive = value
         // setData to empty object because we no longer have a data state.
-        dispatch({ type: 'SET_DATA', payload: {} })
+        // Freeze object to ensure no side effects.
+        dispatch({ type: 'SET_DATA', payload: Object.freeze({}) })
         dispatch({ type: 'SET_FILTERED_DATA', payload: {} })
```
adamdoe, line 362: Does this need to use object.freeze as well?

packages/dashboard/src/helpers/getFilteredData.ts
```diff
@@ -33,7 +33,7 @@ export const getFilteredData = (state: DashboardState, initialFilteredData = {},
 
         newFilteredData[index] = filterData(applicableFilters, formattedData)
       } else {
-        newFilteredData[index] = state.data[row.dataKey] || row.data
+        newFilteredData[index] = state.data[row.dataKey] || row.data || []
```
adamdoe, line 36: nit: outside of scope, but this function has a repeatable pattern that could be abstracted.

adamdoe (approved):
I think this looks good to merge. I didn't test the filtering, but if its looking solid on your end I think we can get it up on a test server for additional testing and revert if needed.

packages/dashboard/src/CdcDashboardComponent.tsx
joshlacey: I couldn't figure out where a side effect was happening and once I froze the object on line 361 I immediately got an error on line 569. Object.freeze is not needed now for the purposes of this pr but I thought it would be good to leave it to prevent any future side effects on Data. I felt like maybe the FilteredData was less permanent? so, IDK if that same reasoning applies to 362. What do you think?

packages/dashboard/src/helpers/getFilteredData.ts
joshlacey: you mean like _data || []

PR #1376, opened 2024-06-24
[DEV-8209] Fix data preview on Dashboard

Atash3000:
Data preview was not showing in some cases  when loading from URL,  it shows if we click other configs and then it shows.Lets see if this version fixes it. Added function inside of effect so it re created function each time dependency changes.

### DIFF
--- a/packages/editor/src/components/PreviewDataTable.tsx
+++ b/packages/editor/src/components/PreviewDataTable.tsx
@@ -71,37 +71,36 @@ const PreviewDataTable = () => {
 
   const dispatch = useContext(EditorDispatchContext)
 
-  const fetchAsyncData = async () => {
-    if (config.type === 'dashboard') {
-      Object.keys(config.datasets).forEach(async datasetKey => {
-        if (config.datasets[datasetKey].preview) {
-          if (config.datasets[datasetKey].dataUrl) {
-            const remoteData = await fetchRemoteData(config.datasets[datasetKey].dataUrl)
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/PreviewDataTable.tsx
```diff
@@ -71,37 +71,36 @@ const PreviewDataTable = () => {
...
-  const fetchAsyncData = async () => {
-    if (config.type === 'dashboard') {
-      Object.keys(config.datasets).forEach(async datasetKey => {
-        if (config.datasets[datasetKey].preview) {
-          if (config.datasets[datasetKey].dataUrl) {
-            const remoteData = await fetchRemoteData(config.datasets[datasetKey].dataUrl)
-            if (Array.isArray(remoteData)) {
-              setTableData(remoteData)
+  useEffect(() => {
+    const fetchAsyncData = async () => {
```
joshlacey, line 75: why is it necessary to move this function into the useEffect?

Atash3000: The function gets re created each time dependancy changes. if it is outside it will not. we have issue on data preview, it does not gets shown and if we trigger other parts of page it comes.Just thought it will be helpfull to fix a bug.Any other ideas ?

joshlacey: I think you could maybe pass in `datasets` into the fetchAsyncData function

PR #1377, opened 2024-06-25
[maps] organize helper functions and add tests

adamdoe:
*Description*
- Moves helper functions to helpers folder and adds starter tests
- `cd packages/maps && yarn test` to run tests


packages/map/src/helpers/generateColorsArray.test.ts
```diff
@@ -0,0 +1,18 @@
+import { generateColorsArray } from '../helpers/generateColorsArray'
```
joshlacey, line 1: could you put this in a "tests" subfolder

packages/map/src/test/hashObj.test.ts
```diff
@@ -0,0 +1,10 @@
+import { hashObj } from '../helpers/hashObj'
```
joshlacey, line 1: move to '../helpers/tests' folder

packages/map/src/test/generateRuntimeLegendHash.test.ts
```diff
@@ -0,0 +1,907 @@
+import { generateRuntimeLegendHash } from '../helpers/generateRuntimeLegendHash'
```
joshlacey, line 1: move to '../helpers/tests' folder

packages/map/src/test/generateRuntimeLegendHash.test.ts
```diff
@@ -0,0 +1,907 @@
...
+        {
+          STATE: 'Wyoming',
+          Rate: 43,
+          Location: 'Vehicle',
+          URL: 'https://www.cdc.gov/'
+        }
+      ],
+      dataFileName: 'valid-data-map.csv',
+      dataFileSourceType: 'file',
+      formattedData: [
```
joshlacey, line 537: can the data be refactored to use fakerjs or to pull the data from some example file

joshlacey (changes requested):

packages/map/src/helpers/generateColorsArray.test.ts
adamdoe: This is the actual function its pulling into the test. Are you suggesting that these functions should be in the test folder as well?

packages/map/src/test/generateRuntimeLegendHash.test.ts
adamdoe: We could use fakerjs, I thought testing a real example after using console logs to get a real hash might be beneficial here to grab the actual hash created. I'm not sure how important that actually is, but we might be able to identify when a real hash has changed/needs to be updated.

packages/map/src/helpers/generateColorsArray.test.ts
joshlacey: yeah create a tests folder in helpers folder and put this generateColorsArray.test.ts file in there. putting the test folder in the folder its testing just reduces clutter for navigating the file tree.

packages/map/src/test/generateRuntimeLegendHash.test.ts
adamdoe: Used example in examples folder for now.

joshlacey (approved):

PR #1379, opened 2024-06-26
[dashboard-preview] Fixed data load function

Atash3000:


packages/editor/src/components/PreviewDataTable.tsx
```diff
@@ -71,37 +71,53 @@ const PreviewDataTable = () => {
...
       }
     }
   }
 
   useEffect(() => {
-    if (!config.data) {
-      fetchAsyncData()
-    } else {
-      setTableData(previewData)
+    const loadData = async () => {
```
joshlacey, line 111: why is this a separate function and why are you awaiting? fetchAsyncData isn't going to block the useEffect so await isn't needed.

joshlacey:
nit: seems like too much abstraction. multiple functions are declared which are only called once. some only have a couple lines of code in them.

PR #1385, opened 2024-07-05
Feature/dev 8667

Atash3000:


adamdoe:
@Atash3000 - please add testing steps here for reviewers.

Atash3000:
Here is Testting steps: Open Line chart. Under Data series "Add Special Line".  for the style select "Filled Circles" which is new added one. Once Filled circles selected you can update the size of the circle below on text field. Also select data series to apply circles.

PR #1389, opened 2024-07-09
[DEV-8698] Applies Filters on MultiDash Switch

mpallansch:


adamdoe (approved):
@mpallansch this lgtm and can be squash/merged

packages/dashboard/src/CdcDashboardComponent.tsx
```diff
@@ -265,6 +265,10 @@ export default function CdcDashboard({ initialState, isEditor = false, isDebug =
     loadAPIFilters(config.dashboard.sharedFilters)
   }, [isEditor, isPreview])
 
+  useEffect(() => {
+    updateDataFilters()
```
joshlacey, line 269: I think if you move this into the first useEffect right after loadAPIFilters it would also work. Did you try that?

joshlacey: I think this is related to this issue you mentioned on 3. in dev-8409

mpallansch: Yes, I had tried that, but the useEffect only has isEditor and isPreview as dependencies, so it doesn't fire when the dashboard multi-tab is changed. I can combine the useEffects possibly, including state.config?.activeDashboard on the first useEffect, but then API filters will reload every time the multi-tab changes.

mpallansch: > I think this is related to this issue you mentioned on 3. in dev-8409

I don't believe so, 3 in 8409 is related to the value not being checked for null in packages/dashboard/src/helpers/generateValuesForFilter.ts

mpallansch: > I think if you move this into the first useEffect right after loadAPIFilters it would also work. Did you try that?

Yes, I had tried that, but the useEffect only has isEditor and isPreview as dependencies, so it doesn't fire when the dashboard multi-tab is changed. I can combine the useEffects possibly, including state.config?.activeDashboard on the first useEffect, but then API filters will reload every time the multi-tab changes.

mpallansch: @joshlacey I combined the useeffects in the latest commit, let me know if looks better now

joshlacey: you should update the shared filters and then pass the result to loadAPIFilters. that's what I'm doing in the dashboard filters refactor branch

packages/dashboard/src/CdcDashboardComponent.tsx
```diff
@@ -340,6 +341,13 @@ export default function CdcDashboard({ initialState, isEditor = false, isDebug =
     dispatch({ type: 'SET_FILTERED_DATA', payload: newFilteredData })
   }
 
+  const updateDataFilters = (sharedFilters = undefined) => {
+    const clonedState = _.cloneDeep(state)
+    if (sharedFilters) clonedState.config.dashboard.sharedFilters = sharedFilters
+    const newFilteredData = getFilteredData(clonedState)
+    dispatch({ type: 'SET_FILTERED_DATA', payload: newFilteredData })
```
joshlacey, line 348: what does the filtered data have to do with the values on the dashboard filter?

PR #1390, opened 2024-07-09
Feature: [dev-8409] dashboard filter refactor

joshlacey:
To configure Dashboard filters you now need to drag and drop a filter widget to the visualizations panel and then select the tools icon to configure.


mpallansch (commented):
Some feedback on testing:

1. Existing dashboards that already have filters still seem to work, but I'm unable to edit the filters in the editor panel
2. The reason the filter-dropdown drag'n'drop element was created in the first place was to repeat filters throughout dashboard, so users didn't have to scroll to the top of long dashboards every time they need to change something. In this setup, they would have to input the filter configuration every time it needed to show and any time they need to change it, they would need to change all instances. 
3. It looks like new filters have a blank option in the dropdown

joshlacey:
@mpallansch 

1. So legacy configuration is supported but I don't think we should support the edit capability beyond what can be done via the Advanced Editor. changes in dev-8574 should make editing easier. What that would entail is user drags over a filter editor and then goes into advanced editor and assigns all of the indexes of the dashboard.shared filters to that new visualization. It should be done this way as I'm not sure how many edge cases we would run into by trying to do this manually.
2. I'm not sure I follow the issue?
3. I'll fix this bug, thanks for catching

mpallansch:
> @mpallansch
> 
> 1. So legacy configuration is supported but I don't think we should support the edit capability beyond what can be done via the Advanced Editor. changes in dev-8574 should make editing easier. What that would entail is user drags over a filter editor and then goes into advanced editor and assigns all of the indexes of the dashboard.shared filters to that new visualization. It should be done this way as I'm not sure how many edge cases we would run into by trying to do this manually.
> 2. I'm not sure I follow the issue?
> 3. I'll fix this bug, thanks for catching

1. How would they go about removing their existing filters to recreate them in the current system? Or would they have to recreate the entire dashboard config if they have filters and need to update them?
2. We can have a meeting to discuss if that's easier, but the need is to be able to configure a dropdown filter a single time, then have it show multiple times on the dashboard

mpallansch (approved):

PR #1406, opened 2024-07-16
[DEV-6410d] Add single series tooltip for Area and Line chart

Atash3000:
Testing Steps:

Select a Line Chart or an Area Chart (STACKED) and add multiple series.
Go to the VISUAL PANEL.
Click "Show hover for single data series."
After the checkbox is ON, hover over the lines in the Line Chart. You should only see the HOVERED series and its corresponding values.
Perform the same test for the stacked Area Chart.


adamdoe (changes requested):
This looks like a great start. I'm seeing the tooltip update properly. It looks like the circles are being removed on hover now though. Can you add those back in?

adamdoe:
Looks good, thanks for the updates!

PR #1411, opened 2024-07-17
Feature/dev 8796

Atash3000:


packages/chart/src/components/BarChart/components/BarChart.Horizontal.tsx
```diff
@@ -75,7 +75,7 @@ export const BarChartHorizontal = () => {
                   const barDefaultLabel = !config.yAxis.displayNumbersOnBar ? '' : yAxisValue
 
                   // check if bar text/value string fits into  each bars.
-                  const textWidth = (getTextWidth as any)(xAxisValue, `normal ${fontSize[config.fontSize]}px sans-serif`)
+                  const textWidth = (getTextWidth as any)(barDefaultLabel, `normal ${fontSize[config.fontSize]}px sans-serif`)
```
adamdoe, line 78: Modify the TS function in `packages/chart/src/types/ChartContext.ts` to support paramaters so you don't have to use this syntax or the any keyword. After modifying the syntax you should be able to use getTextWidth(string, string) when calling the function.

packages/chart/src/components/BarChart/helpers/index.ts
```diff
@@ -76,8 +83,17 @@ export const getBarConfig = ({ bar, defaultBarHeight, defaultBarWidth, config, i
...
+      return label
+    } else {
+      return labelFits && isVertical ? label : !isVertical ? label : ''
+    }
   }
 
   return { barWidthHorizontal, barHeight, isSuppressed, showMissingDataLabel, showZeroValueDataLabel, getBarY, getAbsentDataLabel }
 }
+
+export const testZeroValue = value => {
```
adamdoe, line 96: Remember to write tests for helper functions. This could also be beneficial for other packages and could potentially live in core.

adamdoe (approved):

packages/chart/src/components/BarChart/components/BarChart.Horizontal.tsx
Atash3000: It is modified but still gives red TS error. We need to update TS config main file but i did not do there anything in order not to mess things with configs

PR #1412, opened 2024-07-18
feature [dev-8715] dashboard-multi-select

joshlacey:
testing steps are in the ticket https://websupport.cdc.gov/browse/DEV-8715?filter=-1

### DIFF
--- a/packages/core/helpers/ver/4.24.7.ts
+++ b/packages/core/helpers/ver/4.24.7.ts
@@ -61,6 +61,7 @@ export const dashboardFiltersMigrate = config => {
   // if there's no dashboardFilters visualization but there are sharedFilters create a visualization and update rows.
 
   config.visualizations = newVisualizations
+  delete config.filterBehavior // deprecated
 }
 
 const mapUpdates = newConfig => {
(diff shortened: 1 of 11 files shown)


packages/core/helpers/ver/4.24.7.ts
```diff
@@ -61,6 +61,7 @@ export const dashboardFiltersMigrate = config => {
   // if there's no dashboardFilters visualization but there are sharedFilters create a visualization and update rows.
 
   config.visualizations = newVisualizations
+  delete config.filterBehavior // deprecated
```
adamdoe, line 64: Similar to the active property being removed, as long as this is backwards compatible this is fine.

adamdoe (changes requested):
@joshlacey - This looks good, the only thing I noticed was that the multi select options appear behind visualizations so I think we might need to set a z-index or positioning property that moves the options in front of the visualizations.

packages/core/helpers/ver/4.24.7.ts
joshlacey: I can create my own PR with this. It's basically config cleanup that I forgot to add to the dashboard Filter refactor

PR #1413, opened 2024-07-18
[DEV-8121a] Update font size

Atash3000:


Atash3000:
Updated font size from em to px to enjure child components will not depend on parent size font size and will have own font size 16 px for data table links all over charts/maps etc

PR #1414, opened 2024-07-18
[DEV-8056] Map Patterns > Contrast Checks

adamdoe:


packages/core/components/Alert/components/Alert.tsx
```diff
@@ -0,0 +1,36 @@
...
+  // sanitize the text for setting dangerouslySetInnerHTML
+  const sanitizedData = () => ({
+    __html: DOMPurify.sanitize(message)
+  })
+
+  // reset styles to avoid conflicts in wcms
+  const styleResets = { width: 'unset', height: 'unset', paddingRight: '5px' }
+
+  return (
+    <div className={`alert alert-${type}`} role='alert' style={{ padding: '5px' }}>
```
joshlacey, line 26: does p-1 class not work here?

packages/core/components/Alert/components/Alert.tsx
```diff
@@ -0,0 +1,36 @@
+import { IoMdCheckmarkCircleOutline, IoIosCloseCircle, IoIosInformationCircle } from 'react-icons/io'
```
joshlacey, line 1: why aren't we reusing our Icon component?

packages/map/src/CdcMap.tsx
```diff
@@ -1647,15 +1648,21 @@ const CdcMap = ({ className, config, navigationHandler: customNavigationHandler,
     </a>
   )
 
+  const sectionClassNames = () => {
+    let classes = ['cove-component__content', 'cdc-map-inner-container', `${currentViewport}`]
```
joshlacey, line 1652: const

packages/map/src/components/EditorPanel/components/Panels/Panel.PatternSettings.tsx
```diff
@@ -60,12 +61,23 @@ const PatternSettings = ({ name }: PanelProps) => {
     })
   }
 
+  const checkPatternContrasts = () => {
+    let passesContrast = true
+    state.map.patterns.forEach((pattern, index) => {
```
joshlacey, line 66: nit: hamburger loop can be avoided by using .find()

packages/map/src/components/UsaMap/components/UsaMap.State.tsx
```diff
@@ -118,6 +119,56 @@ const UsaMap = () => {
     }
   }, [data, state.general.territoriesAlwaysShow])
 
+  useEffect(() => {
+    unitedStates.map(geo => {
+      state.map.patterns.map((patternData, patternIndex) => {
+        const { pattern, dataKey, size } = patternData
+        let geoKey = geo.properties.iso
```
joshlacey, line 126: const

packages/map/src/components/UsaMap/components/UsaMap.State.tsx
```diff
@@ -118,6 +119,56 @@ const UsaMap = () => {
...
+        if (!geoKey) return
+        let legendColors
```
joshlacey, line 128: const legendColors = data[geoKey] ? applyLegendToRow(data[geoKey]) : undefined

packages/map/src/components/UsaMap/components/UsaMap.State.tsx
```diff
@@ -118,6 +119,56 @@ const UsaMap = () => {
...
+        const geoData = data[geoKey]
+        if (geoData !== undefined) {
+          legendColors = applyLegendToRow(geoData)
+        }
+        const currentFill = legendColors[0]
+        const hasMatchingValues = patternData.dataValue === geoData[patternData.dataKey]
+        const patternColor = patternData.color || getContrastColor('#000', currentFill)
+
+        if (!hasMatchingValues) return
+        const passesContrastCheck = checkColorContrast(currentFill, patternColor)
```
joshlacey, line 138: fix ordering of statements

```
const hasMatchingValues = patternData.dataValue === geoData[patternData.dataKey] 
if (!hasMatchingValues) return

const currentFill = legendColors[0]
const patternColor = patternData.color || getContrastColor('#000', currentFill)

 const passesContrastCheck = checkColorContrast(currentFill, patternColor)
```

packages/map/src/components/UsaMap/components/UsaMap.State.tsx
```diff
@@ -118,6 +119,56 @@ const UsaMap = () => {
...
+                  }
+                }
+                return pattern
+              })
+            }
+          }
+        })
+      })
+    })
+  }, [state.map.patterns, data, state, setState, applyLegendToRow])
```
joshlacey, line 170: this state.map.patterns dependency seems like a memory leak because of the set state you're calling directly from your useEffect.

once you setState here you have to rerun the useEffect. which leads me to believe that you already had everything you needed set the correct state in the setState context.

joshlacey (changes requested):

packages/core/components/Alert/components/Alert.tsx
adamdoe: Do you see an issue with using react-icons? They provide an easy way to add icons and offer icons that COVE doesn't and it looks like the library is already installed.

packages/map/src/components/UsaMap/components/UsaMap.State.tsx
adamdoe: In state.map.patterns we're checking if the previous state is different on state.map.patterns before calling the setState.

packages/core/components/Alert/components/Alert.tsx
joshlacey: It's another thing we have to maintain... <Icon /> is much more widely used. Does react-icon do anything other than having different icons? We can extend <Icon /> for all of those use cases.

packages/map/src/components/UsaMap/components/UsaMap.State.tsx
```diff
@@ -118,6 +119,54 @@ const UsaMap = () => {
     }
   }, [data, state.general.territoriesAlwaysShow])
 
+  // useEffect(() => {
```
joshlacey, line 122: what's the reason for keeping this?

adamdoe: I removed it on accident, it needs to stay in to keep the pattern contrast checks in place.

packages/core/components/Alert/components/Alert.tsx
adamdoe: I think using a dedicated icon library might be the better choice but I can update this.

joshlacey: I would be fine with using this library if we deprecate the Icon component and add, at a minimum, a comment to it saying to not use it. You'll also need to update the storybook guides that describe how to use Icon component.

packages/map/src/components/UsaMap/components/UsaMap.State.tsx
joshlacey: I feel like you could just remove that from the dependency array and you wouldn't need that check.

packages/core/components/Alert/components/Alert.tsx
adamdoe: I think the Icons we have in COVE are supplemental. For example, chart icons that we want to provide that aren't standard in Icon libraries. Either way, I've gone ahead and updated this component to use the COVE `<Icon />` component now.

packages/map/src/components/UsaMap/components/UsaMap.State.tsx
adamdoe: I've updated the useEffect so that the memory leak isn't happening anymore.

packages/map/src/components/UsaMap/components/UsaMap.State.tsx
```diff
@@ -118,6 +119,77 @@ const UsaMap = () => {
...
+        return {
+          ...prevState,
+          runtime: newRuntime,
+          map: newMap
+        }
+      }
+
+      return prevState
+    })
+  }, [state.map.patterns, data, applyLegendToRow, state.runtime])
```
joshlacey, line 191: Looking at this more, my new question is why is it the child's responsibility to tell the parent it should update its state, when the child actual has no affect on the parent's state. 

I don't a reference to the child state anywhere in this useEfffect. 

It seems like the parent should deal with these updates to map.patterns and runtime before it provides those props to the child.

adamdoe: Thanks for chewing this over. I removed the useEffect in favor of handling the state change inside of the handleGeoUpdate function which I think makes more sense now.

packages/core/components/Alert/components/Alert.tsx
joshlacey: I think the references aren't being used and can be taken out now right?

packages/map/src/components/UsaMap/components/UsaMap.State.tsx
joshlacey: I like the change, much more declarative way of handling it.

packages/core/components/Alert/components/Alert.tsx
adamdoe: Removed the references now.

joshlacey (approved):

PR #1419, opened 2024-07-22
no lerna required

joshlacey:
Gives us the ability to use yarn for everything we are using lerna for. 

Examples: 

`lerna run build` -> `yarn workspaces run build`

`lerna run --scope @cdc/editor start` -> `yarn workspace @cdc/editor run start`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,15 @@
   "version": "0.0.0",
   "private": true,
   "workspaces": [
-    "packages/*"
+    "packages/core",
+    "packages/chart",
+    "packages/map",
+    "packages/filtered-text",
+    "packages/markup-include",
+    "packages/waffle-chart",
+    "packages/data-bite",
+    "packages/dashboard",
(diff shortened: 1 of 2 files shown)


adamdoe (changes requested):
Can you add the commands needed for versioning and publishing to npm/github that `lerna publish` uses? If we're removing lerna from being used we should also remove that package as a dependency.

joshlacey:
I think all the lerna publish and version was doing is running npm under the hood. I'm having trouble really understanding what the difference would be because learna publish isn't working for me.

adamdoe:
I think I'd like to see an interactive way of tagging the packages and publishing them to npm/git with yarn before removing lerna. When useWorkspaces is set to true in lerna.json that functionality is possible with lerna publish.

adamdoe:
Closing until we have solutions for versioning/publishing/tagging that provide similar functionality to lerna publish.

PR #1435, opened 2024-08-01
Dev 8791

mpallansch:
## DEV 8791

Adds option to dynamically generate series

## Testing Steps

Run bar chart, check "Dynamic Series" in series section

## Self Review

## Additional Notes

@adamdoe this is the PR discussed in the meeting that may need to go to the dev environment instead of the test environment


adamdoe (approved):
Tested and looks good.

packages/core/components/DataTable/helpers/getDataSeriesColumns.ts
```diff
@@ -11,8 +11,8 @@ export const getDataSeriesColumns = (config: TableConfig, isVertical: boolean, r
   let tmpSeriesColumns: string[] = []
   if (config.visualizationType !== 'Pie') {
     tmpSeriesColumns = isVertical ? [config.xAxis?.dataKey] : [] //, ...config.runtime.seriesLabelsAll
-    if (config.series) {
-      config.series.forEach(element => {
+    if (config.runtime.series) {
```
joshlacey, line 14: @mpallansch this is breaking. Needs to be config.runtime?.series

packages/chart/src/hooks/useTooltip.tsx
```diff
@@ -96,10 +96,10 @@ export const useTooltip = props => {
 
     const closestXScaleValue = getXValueFromCoordinate(x - Number(config.yAxis.size || 0))
 
-    const includedSeries = visualizationType !== 'Pie' ? config.series.filter(series => series.tooltip === true).map(item => item.dataKey) : config.series.map(item => item.dataKey)
+    const includedSeries = visualizationType !== 'Pie' ? config.runtime.series.filter(series => series.tooltip === true).map(item => item.dataKey) : config.runtime.series.map(item => item.dataKey)
```
joshlacey, line 99: I'm getting an error on dev-8719. note that runtime is not saved to the config so the logic should be able to handle and undefined scenario: https://github.com/CDCgov/cdc-open-viz/blob/495f22439e5c25e05de2ab6c4f44597cd3efce17/packages/core/helpers/formatConfigBeforeSave.ts#L87

mpallansch: Can you send the config that is breaking? In the chart component, the updateConfig function initializes runtime which is called on first load. We had to fix it for standalone DataTables because they may not have a runtime, but charts should always have one

PR #1436, opened 2024-08-01
Feature/dev 8861

adamdoe:
## Self Review


Atash3000 (approved):

packages/core/helpers/coveUpdateWorker.ts
```diff
@@ -14,10 +16,25 @@ export const coveUpdateWorker = config => {
...
 
-  genConfig = update_4_24_3(genConfig)
-  genConfig = update_4_24_4(genConfig)
-  genConfig = update_4_24_5(genConfig)
-  genConfig = update_4_24_7(genConfig)
+  if (!genConfig.version) {
+    genConfig.version = '0.0.1'
+  }
+
+  if (versionNeedsUpdate(genConfig.version, '4.24.3')) {
```
joshlacey, line 23: can you create a helper function for this? like you could create an array of touples. like [versionNumber, versionFunction]

const versions = [['4.24.3', update_4_24_3]...] and loop through that to do the versioning? would be much easier to update.

packages/core/helpers/ver/4.24.9.ts
```diff
@@ -0,0 +1,21 @@
+import _ from 'lodash'
+
+const update_4_24_9 = config => {
+  const ver = '4.24.9'
+
+  const patchSingleStateZoom = newConfig => {
```
joshlacey, line 6: pull this function outside the update function.

packages/core/helpers/ver/4.24.9.ts
```diff
@@ -0,0 +1,21 @@
...
+    // Map zooming is a default feature created for world maps
+    // Since we're turning it on for single state maps lets default it off for opting in.
+    if (newConfig.type === 'map' && newConfig.general.geoType === 'single-state') {
```
joshlacey, line 9: create a helper function outside this function

packages/map/src/components/UsaMap/helpers/map.ts
```diff
@@ -0,0 +1,100 @@
...
+export const getFilterControllingStatePicked = (state, dashboardConfig) => {
+  debugger
+  if (!dashboardConfig?.dashboard?.sharedFilters) {
+    if (state.general.filterControlsStatePicked === '' && !state.general.statePicked.stateName) {
+      return 'Alabama'
+    } else {
+      return state.general.statePicked.stateName
+    }
+  }
+  const filters = dashboardConfig?.dashboard?.sharedFilters
```
joshlacey, line 92: dashboard filters should only filter data, not control visualizations in this way.

joshlacey (changes requested):
@adamdoe I think we should reconsider passing the whole dashboard config down to the visualizations. It seems like the only thing you were referencing from that was the shared filters. I believe still that the shared filters should only act as a data filter and the visualizations should respond only to changes in the data.

joshlacey:
Also I'm not sure that all of those json configs are being used?

packages/map/src/context.ts
```diff
@@ -8,6 +9,7 @@ type MapContext = {
   columnsInData
   currentViewport
   data
+  dashboardConfig: DashboardConfig
```
joshlacey, line 12: remove

packages/dashboard/examples/single-state-filters.json
```diff
@@ -0,0 +1,390 @@
+{
```
joshlacey, line 1: could this go in storybook or should it be attached to the ticket? otherwise delete

packages/map/examples/default-single-state.json
```diff
@@ -1,109 +1,226 @@
 {
```
joshlacey, line 1: could this go in storybook or should it be attached to the ticket? otherwise delete

packages/map/examples/test-single-county-file.json
```diff
@@ -0,0 +1,82 @@
+[
```
joshlacey, line 1: could this go in storybook or should it be attached to the ticket? otherwise delete

joshlacey (changes requested):
I'm not sure what I'm supposed to do with those json configs. how do those help other developers prevent regressions?

joshlacey (approved):

PR #1445, opened 2024-08-05
[DEV-8912] patch shared filters

adamdoe:
## Testing Steps
This patches shared filters. If a shared filter is added without an active value, or migrated without an active value, the first value in the values array is used.

## Self Review

### DIFF
--- a/packages/core/helpers/ver/4.24.9.ts
+++ b/packages/core/helpers/ver/4.24.9.ts
@@ -1,4 +1,5 @@
 import _ from 'lodash'
+import { MultiDashboardConfig } from '@cdc/dashboard/src/types/MultiDashboard'
 
 const patchSingleStateZoom = newConfig => {
   // Map zooming is a default feature created for world maps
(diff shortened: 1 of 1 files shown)


packages/core/helpers/ver/4.24.9.ts
```diff
@@ -8,10 +9,28 @@ const patchSingleStateZoom = newConfig => {
...
 }
 
+const patchDashboardFilterActiveStatus = newConfig => {
+  if (!newConfig.dashboard) return newConfig
+  const dashboardConfig = newConfig as MultiDashboardConfig
+
+  // previous filters need an active status
+  const newSharedFilters = (dashboardConfig.dashboard.sharedFilters || []).map(sf => {
+    if (!sf.active && sf.values) {
+      sf.active = sf?.values?.[0] || 'Select'
```
joshlacey, line 19: values would be defined in this case. as would sf. 'Select' regardless shouldn't be the default. it should be just like null or empty string and you should have set a reset label.

joshlacey (changes requested):

packages/core/helpers/ver/4.24.9.ts
```diff
@@ -8,10 +9,28 @@ const patchSingleStateZoom = newConfig => {
   }
 }
 
+const patchDashboardFilterActiveStatus = newConfig => {
+  if (!newConfig.dashboard) return newConfig
+  const dashboardConfig = newConfig as MultiDashboardConfig
+
+  // previous filters need an active status
+  const newSharedFilters = (dashboardConfig.dashboard.sharedFilters || []).map(sf => {
+    if (!sf.active && sf.values) {
```
joshlacey, line 18: this seems to be reverting the changes I made to formatConfigBeforeSave https://github.com/CDCgov/cdc-open-viz/blob/294572787fe1d47e61a4613636adf640ca30fec0/packages/core/helpers/formatConfigBeforeSave.ts#L57

I added this change because the selection of active state is a side effect of the WCMS developer using the dropdowns. It should instead either be set to some default selection on load of the dashboard either through a query parameter if one is set, to an empty value if reset label is given, or otherwise the first available value.

PR #1452, opened 2024-08-06
Fix [dev-8899] MultiDashboard: tab ordering UI glitches  & <Icon/> updated

joshlacey:
## Testing Steps

1) Create a dashboard.
2) Add visualizations.
3) click "Make Multi-Visualization"
4) Add another tab.
5) Click the left arrow 

Expected: One tab should have visualizations and the other should not.

## Self Review

## Additional Notes

I also fixed the Icon so that we can get type hints for the icon types. I did this thinking I would use an icon on the tabs but ended up not using it. I can pull that change into another PR though if you'd like.

### DIFF
--- a/packages/core/components/ui/Icon.tsx
+++ b/packages/core/components/ui/Icon.tsx
@@ -76,9 +76,25 @@ const iconHash = {
   command: iconCommand
 }
 
+export type IconType = keyof typeof iconHash
+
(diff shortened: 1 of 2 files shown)


adamdoe (approved):
Looks good, thanks for adding the TS additions to Icon as well.

PR #1463, opened 2024-08-15
Hotfix/prettier

adamdoe:
### Discuss
Should we update this to be more manageable now?


joshlacey (approved):
We'll have a lot more changes to look over for a while for files that need to get formatted, but this would make my life easier on my small screens.

PR #1464, opened 2024-08-16
[add-layout-stories] layout stories

adamdoe:
## Self Review


Atash3000:
Legend is not going from "right" to "bottom" when we decrease  screen size. It used to go below chart on certain sizes.

PR #1472, opened 2024-08-21
Feature/dev 8160

adamdoe:
## Testing Steps
1. Open the editor Panel > Filters Accordion
2. Check the box to keep filter styles on mobile
3. Ensure desktop styles are applied on mobile screens

## Self Review


joshlacey:
@adamdoe I'm not seeing anywhere in the ticket that there is a request for a toggle. If anything I would think you should make all future configs default to use the original dropdowns unless on mobile.

The real UX solution here is to make sure each control is mobile friendly. IDK why we're doing all these checks after the fact.

adamdoe:
The original UX ask was to default to dropdowns on mobile. This is so that users can choose to keep the desktop experience.

adamdoe:
inactive

PR #1475, opened 2024-08-23
[type-dimensions] Update dimensions type

Atash3000:
## Testing Steps
Updated types for Dimensions and created separate new type file

## Self Review


packages/core/types/Dimensions.ts
```diff
@@ -0,0 +1 @@
+export type DimensionsType = [string, string]
```
adamdoe, line 1: Should this be a `string | number` for each dimensions?

adamdoe (changes requested):

PR #1493, opened 2024-09-03
Dev-8923 Markup Include Memory Leak

Matt-DiPaolo:
Scenario: In Cove Editor running on your local environment, Select Dashboard, Select some Data set, drag and drop a markup include section onto the visualization panel. Open the Editor.

1) Open the browser console and find the file packages/dashboard/CdcDashboardComponent.
2) put a breakpoint in the updateChildConfig function.

Expected you should not be on an infinite loop.

### DIFF
--- a/packages/markup-include/src/CdcMarkupInclude.tsx
+++ b/packages/markup-include/src/CdcMarkupInclude.tsx
@@ -31,8 +31,21 @@ type CdcMarkupIncludeProps = {
 
 import Title from '@cdc/core/components/ui/Title'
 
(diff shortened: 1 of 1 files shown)


packages/markup-include/src/CdcMarkupInclude.tsx
```diff
@@ -73,8 +80,13 @@ const CdcMarkupInclude: React.FC<CdcMarkupIncludeProps> = ({ configUrl, config:
 
     response.data = responseData
     const processedConfig = { ...(await coveUpdateWorker(response)) }
+    const configNeedsUpdate = !(
```
joshlacey, line 83: I feel like you need to handle this in the useEffect and not in the function itself.

Matt-DiPaolo:
This one has been closed. A different PR is being used.

PR #1495, opened 2024-09-04
dev-9095 SCSS Refactor

joshlacey:
## Testing Steps

Testing on this feature will mostly need to happen through regression. I've run BackstopJS against Storybook and compared Test branch against these changes. Styling wise there were some differences in spacing mostly.

There is a big change of doing a direct import of the Components into Editor and Dashboard. I did this so that we can avoid importing duplicate SCSS. This resulted in about a 50% reduction in the build size of the dashboard and editor packages when you run lerna run build. Also changes to both SCSS and Components now happens without needing to rebuild packages.

## Self Review


adamdoe (approved):
I think this looks good, the only style difference that stood out to me were tooltips on the step one screen. The background is black instead of white now. Nice care with organization and updating of variables. 💯

packages/core/components/ui/Tooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+        className='cove-tooltip--target'
+        data-tooltip-float={float}
+        data-tooltip-place={place}
+        data-tooltip-isOpen={true}
+      >
+        {tooltipTargetChildren ? tooltipTargetChildren.props.children : null}
+      </a>
+      <ReactTooltip
+        id={uid}
+        isOpen={true}
```
joshlacey, line 53: this is causing tooltips to remain open

PR #1497, opened 2024-09-05
fix: dev-9142 Fix missing first grid line

jayb:
This fixes a bug where the first grid line sometimes isn't shown. It happens often with the respiratory virus charts that are [currently live](https://www.cdc.gov/respiratory-viruses/data/activity-levels.html). It would be great if this could go out in the September release.

The bug was introduced in [this commit](https://github.com/CDCgov/cdc-open-viz/pull/1432/commits/8ba391bf64bce964196b9df397bbdf5fd599a864) which was related to DEV-7953.

## Testing Steps

The `testposorig.json` file attatched to the Jira ticket can be used to see the issue this fixes. Load that file on the `dev` branch, see that it's broken, then see that this branch fixes it.


packages/chart/examples/testposorig.json
```diff
@@ -0,0 +1,860 @@
+{
```
joshlacey, line 1: @jayb what is this file being used for?

joshlacey: I would write a ticket in Jira with a description of the bug that you found. You can attach this config in the actual Jira ticket as part of the steps to reproduce the issue. We don't need to save this in our repository.

joshlacey: Once you have the ticket number add the ticket number to the PR. eg: `fix: dev-1234 first grid line`

adamdoe (changes requested):
@jayb 

Nice work! This is working as expected now :)

It looks like this is the feedback to follow:

1. Request a ticket be opened to track this against
2. Remove the example file that was uploaded
3. If you want this included as part of regressions in storybook, move the example file into a story.
4. If you want to practice an interactive rebase you could squash your commits and include the JIRA ticket number in the commit message (otherwise we do this during the merge anyway)

packages/chart/examples/testposorig.json
jayb: I created DEV-9142, moved the test file there and updated the PR title per your recommendation.

jayb:
I did 1&2, but will let you guys do #4 during the merge. Thank you!

PR #1510, opened 2024-09-11
Fix: [dev-9117] url param and multi select

joshlacey:
## Testing Steps

testing instructions are in the ticket

## Self Review

### DIFF
--- a/packages/core/helpers/gatherQueryParams.ts
+++ b/packages/core/helpers/gatherQueryParams.ts
@@ -1,5 +1,11 @@
 import _ from 'lodash'
 
+const isNumber = (value: string) => {
+  const hasLetters = value.match(/[a-zA-Z]/)
+  if (hasLetters) return false
+  return !isNaN(parseInt(value))
+}
+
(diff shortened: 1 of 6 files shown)


packages/core/helpers/gatherQueryParams.ts
```diff
@@ -1,13 +1,19 @@
 import _ from 'lodash'
 
+const isNumber = (value: string) => {
```
adamdoe, line 3: This looks good! I think early on in the project we discussed adding helpers for strings, dates, numbers, etc in this location. `packages/core/helpers/cove/number.js` I'll leave that up to you though.

adamdoe (approved):

packages/core/helpers/gatherQueryParams.ts
joshlacey: This check is something that's really kinda specific to the query parameters and I don't have the confidence that it should be used in all scenarios. I'd make it a helper otherwise.

PR #1518, opened 2024-09-13
[DEV-9171] feat: Extend US Geography

Mgetz10:
**Ticket:** [DEV-9171](https://websupport-cdc.msappproxy.net/browse/DEV-9171)

## Notes
For now I am treating territories as "states" because I have seen a couple data stories that include territories listed in the states dropdown.

I was able to find all the geography for the places listed in the checkForNoData function so I removed the function.

## Testing Steps
Select new territories from state selector dropdown while viewing the default-single-state.json config

## Self Review

## Screenshots (if applicable)
Screenshot 2024-09-13 at 4 57 15 PM


adamdoe (changes requested):
Great job on this! I have a couple of suggestions:

1. We should ensure the "Map Unavailable" message remains in place for unsupported territories.
2. It might be helpful to compare the available territory squares on the state maps to see if there are any additional regions we need to include. (ie. Micronesia, Palau, etc.)

One example might be if someone enters "Micronesia", "San Francisco", or "Foobar" on a dashboard, and we've removed the "Map Unavailable" message, the map will display nothing since those territories aren’t supported. While the territory support seems fine, we should still notify the user when a map isn’t available with the "Map Unavailable" message.

Please review the comments, make the necessary adjustments, and feel free to re-request another pass (sidebar of PR). Thanks and great work!

adamdoe (approved):

PR #1545, opened 2024-09-26
[9293] Hide Null values from hover

Atash3000:
Updated logic for missing data on tooltips. On Stacked bar charts we have checkbox on general panel "Remove Null values from hover" which now removed whole row if value is null ( before it used to remove only value while keeping series name)
this goes to October sprint.

## Self Review

### DIFF
--- a/packages/chart/src/CdcChart.tsx
+++ b/packages/chart/src/CdcChart.tsx
@@ -89,7 +89,7 @@ export default function CdcChart({
   const [loading, setLoading] = useState(true)
   const [colorScale, setColorScale] = useState(null)
   const [config, setConfig] = useState<ChartConfig>({} as ChartConfig)
-  const [stateData, setStateData] = useState(configObj.data || [])
+  const [stateData, setStateData] = useState(configObj?.data || [])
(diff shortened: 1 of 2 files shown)


packages/chart/src/CdcChart.tsx
```diff
@@ -89,7 +89,7 @@ export default function CdcChart({
   const [loading, setLoading] = useState(true)
   const [colorScale, setColorScale] = useState(null)
   const [config, setConfig] = useState<ChartConfig>({} as ChartConfig)
-  const [stateData, setStateData] = useState(configObj.data || [])
+  const [stateData, setStateData] = useState(configObj?.data || [])
```
joshlacey, line 92: configObj should always be defined so we should not add this ? as it passes the error down to the next component in the tree.

joshlacey (changes requested):

packages/chart/src/CdcChart.tsx
adamdoe: I think it might be needed for standalone charts

Atash3000: Fixed!

PR #1555, opened 2024-09-30
Enhancement: [dev-8903] dropdown styling

joshlacey:
## Self Review


packages/core/styles/filters.scss
```diff
@@ -1,11 +1,4 @@
 .filters-section {
-  &__wrapper {
```
adamdoe, line 2: Isn't this needed when titles are above filters? I'm not seeing any space above filters now.

packages/core/components/MultiSelect/multiselect.styles.css
```diff
@@ -1,5 +1,7 @@
 .cove-multiselect {
+  --select-height: calc(1.5em + 0.75rem + 2px);
```
adamdoe, line 2: Can you add a comment for why this calculation is being used?

packages/core/components/MultiSelect/MultiSelect.tsx
```diff
@@ -67,77 +77,86 @@ const MultiSelect: React.FC<MultiSelectProps> = ({ section = null, subsection =
...
+          <div
+            id={multiID}
+            onClick={() => {
+              if (!selectedItems.length) {
+                setExpanded(true)
+              }
             }}
+            className='selected'
           >
-            <Icon display={expanded ? 'caretDown' : 'caretUp'} style={{ cursor: 'pointer' }} />
```
adamdoe, line 107: I'm assuming they don't want the caretUp here anymore for ux reasons.

packages/core/components/ui/Icon.tsx
```diff
@@ -88,15 +88,24 @@ type IconProps = {
   /* Returns icon data as plain svg */
   base?: boolean
   /* Sets alt text for the icon */
-  alt: string
+  alt?: string
```
adamdoe, line 91: We might want to keep this requirement for 508.

packages/core/components/MultiSelect/MultiSelect.tsx
joshlacey: yes that was as per the prototype.

packages/core/components/ui/Icon.tsx
joshlacey: I think having a different description of a checkmark throughout the app might be worse than defaulting to checkmark. I could create a new lookup with default names for each of the icons? so it doesn't end up as 'caretUp'. 

`const defaultNameLookup = { caretUp: 'caret up'}`

`<IconObj title={ alt || defaultNameLookup[display] } />`

if you think this is necessary. I just kept seeing a lot of places where the alt wasn't given and we had typescript errors.

packages/core/styles/filters.scss
joshlacey: Can you screenshot what you're seeing? In Pages/Dashboard/Regressions API Filters Map it looks like there's sufficient space?

adamdoe (approved):

PR #1559, opened 2024-10-01
[DEV-9276] Tighten space below charts

jayb:
## [DEV-9276]

This reduces space below charts on all screen sizes. It does a few main things:

* Removes miscellaneous adjustments that were being added to element heights
* Dynamically positions the xAxis title below the tallest label
* Standardizes spacing of elements below charts to have `20px` of top margin, `0px` of bottom margin, and no top/bottom padding
* Changes default `xAxis.labelOffset` from `65` to `0` and changes all old charts to `0` via `coveUpdateWorker`

Note that changing all old charts to `xAxis.labelOffset=0` may change the x-axis spacing on existing charts a bit, but we think changing everything to `0` is preferable to leaving them at the large value of `65` that most of them have.

## Testing Steps

Look at as many charts as possible and ensure they look good. Everything in Storybook should be no worse than it was before.

## Self Review


packages/chart/src/components/LinearChart.tsx
```diff
@@ -101,7 +103,7 @@ const LinearChart: React.FC<LinearChartProps> = props => {
   let height = config.aspectRatio ? width * config.aspectRatio : config.heights[renderedOrientation]
```
adamdoe, line 103: I think height should remain coming from parentHeight

adamdoe (changes requested):

adamdoe (changes requested):

joshlacey (changes requested):

packages/core/helpers/ver/4.24.10.ts
```diff
@@ -11,9 +12,13 @@ export const removeMultiSelectPropFromMultiselect = newConfig => {
...
 }
 
+export const setXAxisLabelOffsetToZero = newConfig =>
+  editConfigKeys(newConfig, [{ path: ['xAxis', 'labelOffset'], value: 0 }])
+
 const update_4_24_10 = config => {
   const ver = '4.24.10'
-  const newConfig = _.cloneDeep(config)
+  let newConfig = _.cloneDeep(config)
+  newConfig = setXAxisLabelOffsetToZero(config)
```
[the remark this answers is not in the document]
Mgetz10: I was hoping I could start using a pattern of not mutating objects when it makes sense. I like that as a best practice of functional programming because it makes bugs easier for me to trace and the code easier to follow imo. I have a cloneDeep in the editConfigKeys function, so it's baked somewhat into it but I can change it if you object (pun slightly intended). Also game to discuss how we all feel about immutability as a group and rewrite this for now so It matches the existing pattern.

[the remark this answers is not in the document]
Mgetz10: Yeah, there is no real risk here, just a personal preference. I'll mutate for now and hold out for a larger refactor.

packages/chart/src/components/LinearChart.tsx
```diff
@@ -308,13 +323,33 @@ const LinearChart: React.FC<LinearChartProps> = props => {
...
+
+    if (!rightLabel) return
+
+    const axisBottomY = yMax + Number(config.xAxis.axisPadding)
+    const labelRelativeY = rightLabel.getBBox().y - axisBottomY
+    const xLabelY = labelRelativeY + rightLabel.getBBox().height + BOTTOM_LABEL_PADDING
+    setForestXLabelY(xLabelY)
+  }, [config.data.length, forestRowsHeight])
+
+  useLayoutEffect(() => {
```
joshlacey, line 347: https://react.dev/reference/react/useLayoutEffect

packages/chart/src/components/LinearChart.tsx
```diff
@@ -308,13 +323,33 @@ const LinearChart: React.FC<LinearChartProps> = props => {
...
-    setSuffixWidth(textWidth)
+    const suffixEl = suffixRef.current
+    if (!suffixEl && !suffixWidth) return
+    if (!suffixEl) return setSuffixWidth(0)
+    const suffixElWidth = suffixEl.getBBox().width
+    setSuffixWidth(suffixElWidth)
   }, [config.dataFormat.suffix, config.dataFormat.onlyShowTopPrefixSuffix])
 
+  // forest plot x-axis label positioning
+  useEffect(() => {
```
joshlacey, line 334: https://react.dev/learn/you-might-not-need-an-effect

packages/chart/src/test/sizeHelpers.test.ts
```diff
@@ -0,0 +1,80 @@
+import { calcInitialHeight, getOrientation } from '../helpers/sizeHelpers'
```
joshlacey, line 1: can you create a new folder in the helpers folder called 'tests' and place this test there? it's easier to find that way.

packages/core/helpers/ver/4.24.10.ts
```diff
@@ -11,9 +12,15 @@ export const removeMultiSelectPropFromMultiselect = newConfig => {
...
 
+export const setXAxisLabelOffsetToZero = newConfig => {
+  if (!newConfig.xAxis?.labelOffset) return
+  newConfig.xAxis.labelOffset = 0
+}
+
 const update_4_24_10 = config => {
   const ver = '4.24.10'
-  const newConfig = _.cloneDeep(config)
+  let newConfig = _.cloneDeep(config)
```
joshlacey, line 22: revert to const

packages/chart/src/components/LinearChart.tsx
Mgetz10: This was a great read! Definitely converted me to team minimize useEffects.  I think I still need these useEffects here when doing calculations based on the size of rendered elements.  The size of an element isn't available until after it has been rendered to the DOM so unfortunately I think these ones need to stay for now.

Mgetz10:
@joshlacey ready for re-review

packages/chart/src/components/LinearChart.tsx
joshlacey: Why use a state to hold the calculations? the rest of the component doesn't need to know about that, only the element on the page. I'd recommend using a ref and doing a direct DOM manipulation in that case. That way you don't rerender the whole page when you setForestXLabelY.

I did something similar to that here:
https://github.com/CDCgov/cdc-open-viz/pull/1581/files#diff-002175af9bf20550dd9e6e59cfbbf6f86f1986300d62c44289b1aac9ae703c57

Mgetz10: Oh nice! Ok, refactored.

joshlacey (approved):

PR #1583, opened 2024-10-16
Feature/dev 9198

Atash3000:
Testing steps complex: Open Editor Panel then Select Dashboard. On Dashboard add Filters (shared Filters) add Bar chart(any chart would work) on Filters (shared Filters)  add Reset Label.Configre filters Used by:, filter type etc and then Configure chart(dont add local filters to chart).Go to Dashboard Preview: That reset label should be default filter selection when we do Dashboard Preview and also should reset(clear) filters when "reset label" is selected.If need more info please let me know.

## Self Review


packages/dashboard/src/helpers/filterData.ts
```diff
@@ -24,8 +24,9 @@ function getMaxTierAndSetFilterTiers(filters: SharedFilter[]): number {
 }
 
 function filter(data = [], filters: SharedFilter[], condition) {
+  const filteredFilters = filters.filter(f => f.resetLabel !== f.active)
```
adamdoe, line 27: @joshlacey @Atash3000 would activeFilters be a better variable name here?

packages/dashboard/src/helpers/addValuesToDashboardFilters.ts
```diff
@@ -39,7 +43,16 @@ export const addValuesToDashboardFilters = (filters: SharedFilter[], data: Recor
         const active: string[] = Array.isArray(filterCopy.active) ? filterCopy.active : [filterCopy.active]
         filterCopy.active = active.filter(val => defaultValues.includes(val))
       } else {
-        const defaultValue = filterCopy.values[0] || filterCopy.active
+        let hasResetLabel = false
```
adamdoe, line 46: I think resetLabel isn't being set in this code. You could also consider simplifying to:

```.js
const resetLabelFilter = filters.find(fil => fil.resetLabel);
const defaultValue = resetLabelFilter ? resetLabelFilter.resetLabel : filterCopy.values[0] || filterCopy.active;
```

adamdoe (changes requested):

packages/dashboard/src/helpers/filterData.ts
joshlacey: 💯

joshlacey (changes requested):

PR #1585, opened 2024-10-17
Enhancement: Dev-9439 API Error Message Display

Matt-DiPaolo:
## Testing Steps

Scenario: Open a Default Editor and choose the Dashboard Viz. 

Expected: On the Import Data screen

1. Choose Load from URL
2. Add the following link: https://nccd-cove-public-api.apps.ecpaas-dev.cdc.gov/od-public?$datakey=brfss_prevalence_cove_explore_by_t
3. Check the "Always load from URL" checkbox
4. Click on the "Load" button
5. Click the "Configure your Visualization" button

Expected: On the Configure screen

6. Drag a Filter Dropdowns viz into the Visualization area
7. Click on the tools icon

Expected: In the Configure Dashboard Filter screen

8. Click the Filter Accordion
9. Click Add Filter
10. Click the Down arrow button next to "New Dashboard Filter 1"
11. Select URL in the Filter Type dropdown
12. Click inside the textbox of "API EndPoint"
13. Add this "broken" API link (link ends with a "t") : https://nccd-cove-public-api.apps.ecpaas-dev.cdc.gov/od-public?$datakey=brfss_prevalence_cove_explore_by_t
14. Click the "Back to Dashboard" button

Expected back on the Configure screen

Scenario: Clicking on the Dashboard Preview will bring up the preview and the filter will try to pull data from the "broken" link. This will cause an error.

Expected: A red warning box will appear with an error message. This message can be closed if the "X" is clicked on or it will close on it's own in 5 seconds.

## Self Review

- My changes generate no new warnings other than the Error Message in console for Devs

### DIFF
--- a/packages/core/helpers/fetchRemoteData.js
+++ b/packages/core/helpers/fetchRemoteData.js
@@ -1,44 +1,39 @@
 import Papa from 'papaparse'
-import { isSolrCsv, isSolrJson } from '@cdc/core/helpers/isSolr'
+import { isSolrCsv } from '@cdc/core/helpers/isSolr'
 
-export default async function (_url, visualizationType = '') {
+export default function fetchRemoteData(_url) {
   let url = new URL(_url, window.location.origin)
-  try {
-    const path = url.pathname
-    const regex = /(?:\.([^.]+))?$/
-    const ext = regex.exec(path)[1]
+  const path = url.pathname
+  const regex = /(?:\.([^.]+))?$/
+  const ext = regex.exec(path)[1]
 
-    if ('csv' === ext || isSolrCsv(_url)) {
-      return await fetch(url.href)
-        .then(response => response.text())
-        .then(responseText => {
-          // for every comma NOT inside quotes, replace with a pipe delimiter
-          // - this will let commas inside the quotes not be parsed as a new column
-          // - Limitation: if a delimiter other than comma is used in the csv this will break
-          // Examples of other delimiters that would break: tab
-          responseText = responseText.replace(/(".*?")|,/g, (...m) => m[1] || '|')
-          // now strip the double quotes
-          responseText = responseText.replace(/["]+/g, '')
-          const parsedCsv = Papa.parse(responseText, {
-            //quotes: "true",  // dont need these
-            //quoteChar: "'",  // has no effect that I can tell
-            header: true,
-            skipEmptyLines: true,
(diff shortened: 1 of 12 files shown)


joshlacey (changes requested):

packages/dashboard/src/CdcDashboardComponent.tsx
```diff
@@ -167,6 +175,7 @@ export default function CdcDashboard({ initialState, isEditor = false, isDebug =
                 data = transform.autoStandardize(data)
                 data = transform.developerStandardize(data, dataset.dataDescription)
               } catch (e) {
+                dispatchErrorMessages({ type: 'ADD_ERROR_MESSAGE', payload: e })
```
joshlacey, line 178: you don't want to show the actual error message to the user. do:

console.error(e)

then dispatch the generic message to the user, ask UX if you need more guidance on what the message should be.

joshlacey: Also you're missing a catch on the promise chain itself. fetch().then().catch( /* HANDLE ERROR */)

joshlacey (changes requested):

packages/dashboard/src/CdcDashboardComponent.tsx
Matt-DiPaolo: I am not even a 100% sure what the error would be. I am going to put the same one as the other 'There was a problem returning data. Please try again.' for the user.

joshlacey (approved):

packages/core/helpers/fetchRemoteData.js
```diff
@@ -1,44 +1,39 @@
 import Papa from 'papaparse'
-import { isSolrCsv, isSolrJson } from '@cdc/core/helpers/isSolr'
+import { isSolrCsv } from '@cdc/core/helpers/isSolr'
 
-export default async function (_url, visualizationType = '') {
+export default function fetchRemoteData(_url) {
```
joshlacey, line 4: @adamdoe making this whole function async meant that it returns either a promise or a rejected promise. That means in order to handle the error which comes out of this you have to wrap any fetchRemoteData function call with try catch. Rather than doing that I refactored this to a common promise. I'm not sure if this could cause some issues in other components but now I am able to handle errors in a traditional .catch chain. as you see in the CdcDashboardComponent. 
 
 Without this there's no way to handle a 500 error within CdcDashboard.

PR #1600, opened 2024-10-21
[DEV-9187] Fixes Chart Annotations

adamdoe:
## DEV-9187
- Moves svgRef state to CdcChart
- Forwards ref from the rendered svg element to CdcChart so that it can be passed to additional components
- Annotation Panel now uses svgRef to calculate adding the annotations width/height placements

## Testing Steps
- Open a bar, line, or area chart. 
- Click add annotation
- The annotation should appear in the middle of the chart when added

## Self Review

### DIFF
--- a/packages/chart/src/CdcChart.tsx
+++ b/packages/chart/src/CdcChart.tsx
@@ -88,7 +88,7 @@ interface CdcChartProps {
   setSharedFilterValue?: (value: any) => void
   dashboardConfig?: DashboardConfig
 }
-export default function CdcChart({
+const CdcChart = ({
   configUrl,
   config: configObj,
   isEditor = false,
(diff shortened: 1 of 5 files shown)


Mgetz10 (approved):
Works! I noticed that the Stories for "Annotations" don't have a working example. If it's referring to the same functionality maybe we can add a story?

I just tested editing the first story in ChartAnnotation.stories.tsx if you wanted to use that:
`export const Chart_Annotation_Categorical: Story = {
  args: {
    config: editConfigKeys(annotationConfig, [
      {
        path: ['annotations'],
        value: [
          {
            text: 'This is an annotation',
            snapToNearestPoint: false,
            fontSize: 16,
            bezier: 10,
            show: {
              desktop: true,
              tablet: true,
              mobile: true
            },
            connectorType: 'line',
            colors: {
              label: 'black',
              connector: 'black',
              marker: 'black'
            },
            selected: true,
            anchor: {
              vertical: false,
              horizontal: false
            },
            marker: 'arrow',
            edit: {
              subject: true,
              label: true
            },
            seriesKey: '',
            x: 50,
            y: 194.92950439453125,
            xKey: '1/15/2016',
            yKey: '',
            dx: 108,
            dy: -86,
            opacity: '100',
            savedDimensions: [null, null],
            connectionType: 'line'
          }
        ]
      }
    ])
  }
}`

PR #1601, opened 2024-10-21
fix [dev-9566] table columns sizing

joshlacey:

Table columns were collapsing with asterisks to the right when on mobile, this fixes that an makes it scroll instead of collapsing. 

## Self Review


jayb (approved):
Looks great, thank you.

PR #1603, opened 2024-10-22
[DEV-9293a] Update tooltips for null values

Atash3000:
Updated tooltips for bar charts to not to shows null or undefined values when user hovers. To test it we need null or missing data in dataset and then in General Panel checkox "Remove Null values from hover". it removes nulls.

## Self Review


packages/chart/src/hooks/useTooltip.tsx
```diff
@@ -198,8 +195,16 @@ export const useTooltip = props => {
             )
             ?.flatMap(seriesKey => {
               const value = resolvedScaleValues[0]?.[seriesKey]
+              console.log(value, 'value')
```
adamdoe, line 198: remove

adamdoe (changes requested):

PR #1610, opened 2024-10-24
FIX: DEV-9346 Nested Dropdown: not populating filter options (Viz level)

Matt-DiPaolo:
## DEV-9346 Nested Dropdown: not populating filter options (Viz level)
[https://websupport.cdc.gov/browse/DEV-9346](https://websupport.cdc.gov/browse/DEV-9346)

When a nested dropdown option is selected and parameters are used, the URL will change showing the newly selected option. 

Also, when parameters are used, using the URL with parameters will selected the nested dropdown option and display the correct chart on the screen  

## Testing Steps

Scenario: Upload [dev-9346-nested-dropdown-using-parameters.json](https://github.com/user-attachments/files/17509307/dev-9346-nested-dropdown-using-parameters.json), Click on Dashboard Preview 

Expected: There is a bar chart with a nested dropdown filter.  

1. Click on the Nested Dropdown and select Age Group 3 - Overall
2. Once loaded, copy the URL
3. Click on the Nested Dropdown and select Age Group - 50-64 years
4. Replace current URL with the copied URL from step 2 
5. Go to that URL

Expected: The Chart from step 3 will show Age Group - 50-64 years has 6 bars. After navigating the the URL in step 2 there will be 1 single bar and the nested dropdown will read "Age Group 3 - Overall"

## Self Review

### DIFF
--- a/packages/chart/src/CdcChart.tsx
+++ b/packages/chart/src/CdcChart.tsx
@@ -73,6 +73,7 @@ import SkipTo from '@cdc/core/components/elements/SkipTo'
 import { filterVizData } from '@cdc/core/helpers/filterVizData'
 import LegendWrapper from './components/LegendWrapper'
 import _ from 'lodash'
+import { addValuesToFilters } from '@cdc/core/helpers/addValuesToFilters'
 
 interface CdcChartProps {
   configUrl?: string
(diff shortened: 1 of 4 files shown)


joshlacey (approved):

packages/chart/src/CdcChart.tsx
```diff
@@ -373,28 +366,8 @@ const CdcChart = ({
...
-        newConfig.filters[index].active =
-          !newConfig.filters[index].active || !includes(filterValues, newConfig.filters[index].active)
-            ? filterValues[0]
-            : newConfig.filters[index].active
-        newConfig.filters[index].filterStyle = newConfig.filters[index].filterStyle
-          ? newConfig.filters[index].filterStyle
-          : 'dropdown'
-      })
-      currentData = filterVizData(newConfig.filters, newExcludedData)
+      const filtersWithValues = addValuesToFilters(newConfig.filters, newExcludedData)
```
joshlacey, line 369: @adamdoe addValuesToFilter seemed to be doing the same thing as what we were trying to do here. We saw that before this refactor the app wasn't doing everythign it needed to do with query parameters and such. addValuesToFilter should fix that.

We weren't sure how to make these same changes to maps... not sure if other packages may need a similar change.

PR #1617, opened 2024-10-28
Enhancement: Dev-8931 Hide Markup Include on null values

Matt-DiPaolo:
## Enhancement: Dev-8931 Hide Markup Include on null values
[https://websupport.cdc.gov/projects/DEV/issues/DEV-8931](https://websupport.cdc.gov/projects/DEV/issues/DEV-8931)

Added ability to display/hide markup include depending on if there are values with null values.

## Testing Steps

Testing Instructions

Scenario: Upload [dev-8931-hide-markup-include.json](https://github.com/user-attachments/files/17544291/dev-8931-hide-markup-include.json), Click on Dashboard Preview 
Expected: There is a Dashboard Filter and a Markup Include with text

1. Choose Normal Weight from the dropdown.
Expected: Markup Include is shown with no value next to Percentage(s).

2. Choose Not Checked in past 5 Years
Expected: Markup Include is shown with no value next to Class(es).

3. Click Dashboard Description
4. Click the tools icon on the Markup Include
5. Open Content Editor 
6. Click the checkbox next to Hide section on null
7. Open Dashboard Preview

Expected: When Normal Weight or Not Checked in past 5 years is chosen, there shouldn't be any Markup Include showing. If you choose any other option that has values for both variables then the Markup Include will show. 

## Self Review

## Additional Notes

In the EditorPanel.tsx file, the only changes I made are from 199-221. The rest of the changes are linter changes.


packages/markup-include/src/CdcMarkupInclude.tsx
```diff
@@ -58,6 +58,12 @@ const CdcMarkupInclude: React.FC<CdcMarkupIncludeProps> = ({
   const { contentEditor, theme } = config || {}
   const data = configObj?.data
 
+  const variablesCheck = useMemo(() => {
```
joshlacey, line 61: we'll need to find another way to handle this.

packages/markup-include/src/CdcMarkupInclude.tsx
```diff
@@ -195,6 +206,12 @@ const CdcMarkupInclude: React.FC<CdcMarkupIncludeProps> = ({
         finalDisplay = finalDisplay.toLocaleString('en-US', { useGrouping: true })
       }
 
+      if (finalDisplay === '' && contentEditor.allowHideSection) {
+        variablesCheck.isEmpty = true
```
joshlacey, line 210: don't directly set a stateful object. this will cause side effects

packages/markup-include/src/CdcMarkupInclude.tsx
joshlacey: The correct thing to do would be to search the data given the configuration you have for null values. It seems like you have a function below that does something like that, I wouldn't use that though. I think if you need to abstract some of the functions that are in that below function you should do that so they can be reused in both places.

joshlacey (changes requested):

Matt-DiPaolo:
I also changed the code that was adding commas to the variable. The old code was only displaying one value that had the comma and not the whole array. This allowed me to remove the let variable, too.

joshlacey (approved):

PR #1629, opened 2024-10-30
Feature/dev 9630

Atash3000:
Update scales for Date-time and Date type of x axis. 
1) Open Bar/Line/Combo chart. 
2) Date Category Panel ==> select scaling  type "Date(Linear)" or "Date(Time) scale.
2) Setup date format on config.
3) below  "Data scaling type " checkbox there is new checkbox " Show Dates newest to oldest"
4) By clicking checkbox the bars along with ticks should reverse from old to new vise vers.
5)Perform same test on Combo/Line.

## Self Review


packages/chart/src/components/EditorPanel/EditorPanel.tsx
```diff
@@ -2517,14 +2517,20 @@ const EditorPanel = () => {
...
                           </label>
-
                           <CheckBox
                             value={config.xAxis.manual}
                             section='xAxis'
                             fieldName='manual'
                             label='Manual Ticks'
                             updateField={updateField}
                           />
+                          <CheckBox
```
adamdoe, line 2527: Only show this setting when the scale type is a date scale.

adamdoe (changes requested):
Looking good, lets just adjust that setting so its only visible when the scale type is date time or date linear. Thanks!

packages/chart/src/components/EditorPanel/EditorPanel.tsx
Atash3000: Fixed!

PR #1638, opened 2024-10-31
[DEV-9182] Align image button to the right

Atash3000:
Open Data table panel ==>click "Show Download Image" the button has to be always on the right side of chart.
We had bug that it is on the left sometimes.

## Self Review

### DIFF
--- a/packages/core/styles/_button-section.scss
+++ b/packages/core/styles/_button-section.scss
@@ -2,8 +2,8 @@
 .download-buttons {
   display: flex;
   justify-content: flex-end;
-  margin: 0 1em; // align w/ data table
-  margin-left: auto;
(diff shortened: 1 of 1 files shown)


packages/core/styles/_button-section.scss
```diff
@@ -3,7 +3,7 @@
   display: flex;
   justify-content: flex-end;
   margin: 0 1em; // align w/ data table
-
+  margin-left: auto;
```
adamdoe, line 6: Consolidate with the top style and I believe we can remove the 1em. Test:

`margin: 0 0 0 auto`

adamdoe (changes requested):
See style feedback

PR #1639, opened 2024-10-31
[DEV-9555] Fix initial height of horizontal bar charts in editor

jayb:
## [DEV-9555]

See the ticket; the height of horizontal bar charts is initially wrong when created in the editor, and doesn't correct itself until another option is toggled.

## Testing Steps

1. Run the editor locally
2. Create a bar chart with the sample data set, change orientation to horizontal
3. Click "I'm done"
4. See that the chart is the correct height and isn't cut off by the data table on this branch

## Self Review


Mgetz10 (approved):
Works!

PR #1651, opened 2024-11-05
Feature/dev 9503

Atash3000:
## Testing Steps
Open Box plot, make you you have data set with multiseries.
1. We need to add more series to see results where box plot chart can render more series. used to be one only.
2. Box plot tooltip issue fixed when user enters custom measure value like "Maximum"
3. Updated "Total" measure into "Count" measure where we show  count of items instead of sum.

## Self Review


packages/chart/src/CdcChart.tsx
```diff
@@ -438,20 +435,11 @@ const CdcChart = ({
...
-
-      const uniqueArray = function (arrArg) {
-        return arrArg.filter(function (elem, pos, arr) {
-          return arr.indexOf(elem) === pos
-        })
-      }
+      const combinedData = filteredData || data
+      let allKeys = combinedData.map(d => d[newConfig.xAxis.dataKey])
+      const groups = _.uniq(allKeys)
+      console.log(groups, 'groups')
```
adamdoe, line 441: remove

packages/chart/src/components/EditorPanel/EditorPanel.tsx
```diff
@@ -816,19 +816,6 @@ const EditorPanel = () => {
     return null
   }
 
-  useEffect(() => {
```
adamdoe, line 819: good!

packages/chart/src/components/LinearChart.tsx
```diff
@@ -41,6 +41,7 @@ import { useTooltip as useCoveTooltip } from '../hooks/useTooltip'
 import { useEditorPermissions } from './EditorPanel/useEditorPermissions'
 import Annotation from './Annotations'
 import { BlurStrokeText } from '@cdc/core/components/BlurStrokeText'
+import BoxPlotChart from './BoxPlot/BoxPlotX'
```
adamdoe, line 44: This import doesn't seem correct

packages/chart/src/components/BoxPlot/BoxPlot.tsx
```diff
@@ -0,0 +1,189 @@
+import React, { useContext } from 'react'
```
adamdoe, line 1: What is new in this file?

packages/chart/src/data/initial-state.js
```diff
@@ -68,9 +68,6 @@ export default {
   boxplot: {
     plots: [],
     borders: 'true',
-    firstQuartilePercentage: 25,
```
adamdoe, line 71: It looks like we can remove these from ts as well

adamdoe (changes requested):
Can you add a story or example file here? When I plot price I'm not seeing additional boxes added to the chart area. I see some changes but I'm not sure if its correct. I added a few additional comments for edits. Feel free to closes those after making updates. Thanks, and great work as always!

packages/chart/src/components/BoxPlot/BoxPlot.tsx
Atash3000: New box plot added.

packages/chart/src/components/LinearChart.tsx
Atash3000: removed

packages/chart/src/CdcChart.tsx
Atash3000: removed

packages/chart/src/data/initial-state.js
Atash3000: removed

adamdoe (changes requested):
In the storybook example, are you able to include outliers/non outliers so we can see the full example?

PR #1655, opened 2024-11-06
 [DEV-9726] Map Background & Border Color Updates

adamdoe:
## [DEV-9726] Map Background & Border Color Updates
- Globally updates map border and background colors
- Add types to InputToggle and convert from jsx to tsx
- Update Alert component to allow no dismiss button
- Add cool-gray color palette options to global styles file
- Add version type

TODO: removeVersionNeedsUpdate in CdcMap.tsx

## Self Review


packages/map/src/CdcMap.tsx
```diff
@@ -394,9 +395,13 @@ const CdcMap = ({
 
       // Special Classes (No Data)
       if (result[legendIdx].special) {
-        const specialClassColors = chroma.scale(['#D4D4D4', '#939393']).colors(specialClasses)
-
-        return specialClassColors[legendIdx]
+        if (versionNeedsUpdate(state.version, '4.24.11')) {
```
joshlacey, line 398: @adamdoe this is a big nit, but it's kinda confusing that this says versionNeedsUpdate. like that to me implies that you're going to do something to update it. could we possible alias that function as something more obvious? like maybe `needsLegacySupport(state.version, '4.24.11')`

joshlacey: like we could export another function from the versionNeedsUpdate function that is just this alias, that way others will know the other name.

we could pick a more universal name too like `isOlderVersion`

We could do this in a totally different PR too, just wanted to point it out here.

packages/map/src/types/MapConfig.ts
```diff
@@ -164,4 +164,6 @@ export type MapConfig = Visualization & {
   hexMap: HexMapSettings
   filterBehavior: string
   visual: MapVisualSettings
+  // version of the map
+  version: `${number}.${number}.${number}`
```
joshlacey, line 168: Can we move this out to the core library and import it? everything will be using this

`type VersionNumber = \`${number}.${number}.${number}\``

joshlacey (approved):

PR #1659, opened 2024-11-07
[DEV-9561] Fix top bottom legend default border

Atash3000:
Fixed legend border when legend changes from side to top/bottom the border around legend should be turn of by default.
there is a checkbox "Hide Legend Box" which controls it.

## Self Review

### DIFF
--- a/packages/core/helpers/useDataVizClasses.ts
+++ b/packages/core/helpers/useDataVizClasses.ts
@@ -67,16 +67,20 @@ export default function useDataVizClasses(config, viewport = null) {
     if (legend?.style === 'gradient') ulClasses.push('patterns-only')
     return ulClasses
   }
-  const hasBorder = legend?.hideBorder ? 'no-border' : ''
+  const hasSideBorder = legend.position === 'side' && legend?.hideSideBorder ? 'no-side-border' : ''
+  const hasTopBottomBorder = legend?.hideTopBottomBorder && legend.position !== 'side' ? 'no-top-bottom-border' : ''
   const legendOuterClasses = [
     `${legend?.position}`,
     `${getListPosition()}`,
     `cdcdataviz-sr-focusable`,
     `${viewport}`,
-    `${hasBorder}`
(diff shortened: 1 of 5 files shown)


packages/map/src/components/Legend/components/index.scss
```diff
@@ -52,7 +52,11 @@
           flex-wrap: wrap;
         }
       }
-      &.no-border {
+      &.no-side-border {
```
adamdoe, line 55: Use bootstrap `border-0` class:
https://getbootstrap.com/docs/4.0/utilities/borders/

packages/map/src/components/Legend/components/index.scss
```diff
@@ -52,7 +52,11 @@
...
+        border: none;
+      }
+
+      &.no-top-bottom-border {
```
adamdoe, line 59: Use bootstrap `border-top-0` class

packages/core/helpers/useDataVizClasses.ts
```diff
@@ -67,16 +67,20 @@ export default function useDataVizClasses(config, viewport = null) {
     if (legend?.style === 'gradient') ulClasses.push('patterns-only')
     return ulClasses
   }
-  const hasBorder = legend?.hideBorder ? 'no-border' : ''
+  const hasSideBorder = legend.position === 'side' && legend?.hideSideBorder ? 'no-side-border' : ''
+  const hasTopBottomBorder = legend?.hideTopBottomBorder && legend.position !== 'side' ? 'no-top-bottom-border' : ''
```
adamdoe, line 71: Can we streamline the logic and user interactions? For instance, could we implement a single check for the legend borders? If the user selects the bottom legend, we’d show a specific border style, and if the legend is positioned on the side, we’d display a different border style without having to bloat the initial state file with different positions. 

Let me know if that makes sense or if its something custom thats needed here.

adamdoe (changes requested):

PR #1722, opened 2024-11-26
Enhancement: auto calculate padding for overflowing grid labels

Mgetz10:
## No Ticket

In an effort to never overlap data with grid when "Only show top prefix/suffix" is activated - auto generate padding if gridline is above maxValue

## Testing Steps
1. Create of view linear chart vertical chart with "Only show top prefix/suffix" activated
2. Observe the label does not overlap data

## Self Review

### Before
Screenshot 2024-11-26 at 2 40 04 PM

### After
Screenshot 2024-11-26 at 2 40 11 PM

### DIFF
--- a/packages/chart/src/components/LinearChart.tsx
+++ b/packages/chart/src/components/LinearChart.tsx
@@ -175,7 +175,6 @@ const LinearChart = forwardRef<SVGAElement, LinearChartProps>(({ parentHeight, p
   }
 
   // VARS/MEMOS
-  const labelsOverflow = onlyShowTopPrefixSuffix || labelsAboveGridlines
   const shouldAbbreviate = true
   const isHorizontal = orientation === 'horizontal' || config.visualizationType === 'Forest Plot'
   const isLogarithmicAxis = config.yAxis.type === 'logarithmic'
@@ -251,8 +250,8 @@ const LinearChart = forwardRef<SVGAElement, LinearChartProps>(({ parentHeight, p
       ...config,
       yAxis: {
         ...config.yAxis,
-        scalePadding: labelsOverflow ? autoPadding : config.yAxis.scalePadding,
-        enablePadding: labelsOverflow || config.yAxis.enablePadding
+        scalePadding: onlyShowTopPrefixSuffix ? autoPadding : config.yAxis.scalePadding,
+        enablePadding: onlyShowTopPrefixSuffix || config.yAxis.enablePadding
       }
     },
     minValue,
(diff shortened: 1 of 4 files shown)


packages/chart/src/components/LinearChart.tsx
```diff
@@ -100,6 +100,7 @@ const LinearChart = forwardRef<SVGAElement, LinearChartProps>(({ parentHeight, p
   const [animatedChart, setAnimatedChart] = useState(false)
   const [point, setPoint] = useState({ x: 0, y: 0 })
   const [suffixWidth, setSuffixWidth] = useState(0)
+  const [autoPadding, setAutoPadding] = useState(0)
```
adamdoe, line 103: I hate to make this name longer, but I wonder if we should explicitly state what type of autoPadding this is (ie. yAxisAutoPadding) should we introduce other automatic padding into the mix.

packages/chart/src/components/LinearChart.tsx
```diff
@@ -114,12 +115,73 @@ const LinearChart = forwardRef<SVGAElement, LinearChartProps>(({ parentHeight, p
     freezeOnceVisible: false
   })
 
+  // Pre render functions
+  const countNumOfTicks = (axis, max) => {
```
adamdoe, line 119: nit: Can we move this into `/helpers/countNumOfTicks.ts`?

packages/chart/src/components/LinearChart.tsx
```diff
@@ -114,12 +115,73 @@ const LinearChart = forwardRef<SVGAElement, LinearChartProps>(({ parentHeight, p
...
+  }
+
   // VARS/MEMOS
   const shouldAbbreviate = true
   const isHorizontal = orientation === 'horizontal' || config.visualizationType === 'Forest Plot'
   const isLogarithmicAxis = config.yAxis.type === 'logarithmic'
   const isForestPlot = visualizationType === 'Forest Plot'
   const suffixHasNoSpace = !suffix.includes(' ')
+  const padding = orientation === 'horizontal' ? Number(config.xAxis.size) : Number(config.yAxis.size)
+  const fontSize = { small: 16, medium: 18, large: 20 }
```
adamdoe, line 184: Can we remove this and use `packages/core/helpers/cove/fontSettings.ts`. I think this is one of those repeated values we need to remove and reducing the amount of repetition or removing it altogether in favor of CSS (if possible) would be helpful.

adamdoe (changes requested):

packages/chart/src/components/LinearChart.tsx
Mgetz10: @adamdoe is ok if we keep this here and save the refactor for a more relevant change? Its not really related to work i did - i just moved it up from a lower part of the component in an attempt to keep the variables together

adamdoe: Sure, thats fine!

Mgetz10: @adamdoe Oh sorry I didn't realize how easy the request was! adding in the change 🫡

Mgetz10:
@adamdoe @jayb Ready for review

Mgetz10:
@adamdoe NVM! new bug found

jayb (approved):
Works great!

Mgetz10:
@adamdoe ok ready for you when you have a moment

adamdoe (approved):

PR #1747, opened 2024-12-06
Fix [dev-9970] dashboard long page load time

joshlacey:
## Testing Steps

testing steps are in the ticket and require connection to Chronic API

## Self Review

### DIFF
--- a/packages/core/components/Legend/Legend.Gradient.tsx
+++ b/packages/core/components/Legend/Legend.Gradient.tsx
@@ -38,7 +38,7 @@ const LegendGradient = ({
 
   const numTicks = colors?.length
 
-  const longestLabel = labels && labels.length > 0 ? labels.reduce((a, b) => (a.length > b.length ? a : b)) : ''
+  const longestLabel = (labels || []).reduce((a: string, b) => (a.length > String(b).length ? a : b), '')
   const boxHeight = 20
   let height = 50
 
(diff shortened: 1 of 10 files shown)


packages/core/components/Legend/Legend.Gradient.tsx
```diff
@@ -38,7 +38,7 @@ const LegendGradient = ({
 
   const numTicks = colors?.length
 
-  const longestLabel = labels && labels.length > 0 ? labels.reduce((a, b) => (a.length > b.length ? a : b)) : ''
+  const longestLabel = (labels || []).reduce((a: string, b) => (a.length > String(b).length ? a : b), '')
```
joshlacey, line 41: @Atash3000 @adamdoe  I made this change to fix an issue one of the wcms devs was seeing. labels was coming back from the api with a null value in the array and null.length isn't a function. stringifying it makes sure it will at least show instead of breaking COVE

Matt-DiPaolo:
1. When the page is originally loaded, I don't see the Markup Include section having any quarters showing. 

2. `?cove-auto-load=true` is working properly and is showing up in the Markup Include

3. `?Quarter=2024-Q4` is working properly but I would suggest changing it to `?Quarter=2022-Q1` or something else so you can see the difference. "2024-Q4" is the default.

Matt-DiPaolo (approved):
Everything is working properly.

Atash3000 (approved):

PR #1761, opened 2024-12-10
FIX: [DEV-9924a]-Box Plot: Table Issues

Atash3000:
## Testing Steps
1.Updated Box plot max calculation logic. 
2.Added unique key to JSX return.
Fixed small bugs.

## Self Review


packages/chart/src/components/BoxPlot/BoxPlot.tsx
```diff
@@ -5,6 +5,7 @@ import ConfigContext from '../../ConfigContext'
 import ErrorBoundary from '@cdc/core/components/ErrorBoundary'
 import { colorPalettesChart } from '@cdc/core/data/colorPalettes'
 import { handleTooltip, calculateBoxPlotStats, createPlots } from './helpers/index'
+import { v4 as uuidv4 } from 'uuid'
```
adamdoe, line 8: Is this needed in the file or was it previously removed?

Atash3000: Removed. We don't need it.

PR #1792, opened 2024-12-19
Enhancement [DEV-10065] conform legends to ux specs

Mgetz10:
## [DEV-10065](https://websupport-cdc.msappproxy.net/browse/DEV-10065)

Conform spacing and font sizes to [Fonts, Spacing, and Elements in Charts [UX-1249]](https://uxpreview.cdc.gov/3GE5EW/?id=6cg4ol&p=fonts__spacing__and_elements_in_charts__ux-1249_&sc=3&g=15)

## Testing Steps

Inspect and compare legends to UX document

## Self Review


.storybook/preview-head.html
```diff
@@ -1,26 +1,27 @@
...
+  }
 </style>
-<link rel="stylesheet prefetch" href="https://www.cdc.gov/TemplatePackage/contrib/libs/bootstrap/latest/css/bootstrap.min.css?_=39423">
-<link rel="stylesheet prefetch" href="https://www.cdc.gov/TemplatePackage/4.0/assets/css/app.min.css?_=39423">
-
-<!-- to toggle on tp5 -->
-<!-- <link rel="stylesheet prefetch" href="https://wcms-wp-test.cdc.gov/TemplatePackage/5.0/css/app.css?_=81698"> -->
+<link
+  rel="stylesheet prefetch"
+  href="https://www.cdc.gov/TemplatePackage/contrib/libs/bootstrap/latest/css/bootstrap.min.css?_=39423"
```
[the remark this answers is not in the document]
Mgetz10: I think you're right! I removed all the bootstrap imports

.storybook/coveSbThemes.js
```diff
@@ -19,7 +19,7 @@ export const managerTheme = create({
   appBorderRadius: 0,
 
   // Typography
-  fontBase: '"Open Sans", sans-serif',
+  fontBase: '"Nunito", sans-serif',
```
adamdoe, line 22: nit: in this file I'm wondering if we could remove the typography settings in favor of importing CSS. If its overly complicated this is fine, especially since its Storybook.
nit: convert previous js files to ts files when possible.

packages/chart/src/scss/main.scss
```diff
@@ -153,23 +153,46 @@
...
     vertical-align: top;
     text-align: left;
     border: 1px solid var(--lightGray);
     position: relative;
 
+    h3,
+    p,
+    tspan,
+    div {
+      // TODO: when we have updated each section to use the new font, we can transition to using a global style
```
adamdoe, line 165: Would you rather keep this here or transition it into core? .chart-container { ...styles }. I'll let you make that call, but if we're working across different packages you might consider moving it so other developers see it across the different packages.

packages/core/components/Legend/Legend.Gradient.tsx
```diff
@@ -92,10 +92,7 @@ const LegendGradient = ({
 
   if (style === 'gradient') {
     return (
-      <svg
-        style={{ overflow: 'visible', width: '100%', marginTop: 10, marginBottom: hideBorder ? 10 : 0 }}
-        height={newHeight}
-      >
+      <svg style={{ overflow: 'visible', width: '100%' }} height={newHeight}>
```
adamdoe, line 95: nit: w-100, overflow-visible as classes.

packages/map/src/components/Legend/components/index.scss
```diff
@@ -70,33 +67,44 @@
     }
 
     .legend-container {
-      --space-between-legend-items: 0.6em;
       position: relative;
-      &.legend-padding {
```
adamdoe, line 75: Similar to the comment made in charts, you could move these sections into core so that its hopefully easier to see the containers your targeting. I realize there are differences on the divs, but it might help normalize them going forward.

adamdoe (changes requested):
Nice work. I like the amount of deletions in CSS and the start of normalizing these fonts using CSS variables. A few of the comments made are likely from outdated code, and just used as reminders while working in the files for slow refactoring. Only nits here, nice work!

packages/chart/src/scss/main.scss
Mgetz10: great idea

.storybook/coveSbThemes.js
Mgetz10: Partially resolved: I changed this file to typescript but so far I wasn't able to figure out a theming approach using imported css - going to put that enhancement off for now in favor of getting this pr in so we can begin other UX tickets

adamdoe (approved):

PR #1820, opened 2025-01-08
FIX: Feature-[DEV-10046] : Dashboard Filters - Parent

Atash3000:
## Testing Steps
Set Up Filters in the Dashboard:

Create at least two filters with "Data" as the filter type.
Configure one as a Parent Filter and the other as a Child Filter (link the child to the parent using the Parents field).
Test Parent-Child Logic:

Select a value in the Parent Filter (e.g., Electronics).
Verify that the Child Filter updates to display only values related to the selected parent value (e.g., Sony, Samsung).
Screenshot 2025-01-08 at 14 22 45

## Self Review


adamdoe:
@Atash3000 - It looks like there are failing unit tests on this PR.

Atash3000:
@adamdoe  updated unit test. Should be good now.

packages/dashboard/src/helpers/updateChildFilters.ts
```diff
@@ -0,0 +1,43 @@
+import { SharedFilter } from '../types/SharedFilter'
+import _ from 'lodash'
+
+export const updateChildFilters = (newSharedFilters: SharedFilter[], data: Record<string, any>): SharedFilter[] => {
+  const dateSet = Object.values(data).flat()
```
adamdoe, line 5: spelling

packages/dashboard/src/helpers/updateChildFilters.ts
```diff
@@ -0,0 +1,43 @@
...
+  const updatedFilters = _.cloneDeep(newSharedFilters)
+
+  // Update each child filter
+  childFilterIndexes.forEach(childIndex => {
+    const childFilter = newSharedFilters[childIndex]
+    const parentFilter = newSharedFilters.find(filter => String(childFilter.parents) === String(filter.key))
+
+    if (parentFilter) {
+      // Filter dataset based on parent's active value
+      const parentActiveValuesArr = dateSet.filter(
```
adamdoe, line 24: spelling

adamdoe (changes requested):

packages/dashboard/src/helpers/updateChildFilters.ts
Atash3000: Fixed!

packages/dashboard/src/helpers/updateChildFilters.ts
Atash3000: Fixed!

joshlacey:
@Atash3000 The parent filter selector should be a multi select.

joshlacey (approved):

packages/dashboard/src/helpers/updateChildFilters.ts
```diff
@@ -0,0 +1,44 @@
+import { SharedFilter } from '../types/SharedFilter'
+import _ from 'lodash'
+
+export const updateChildFilters = (newSharedFilters: SharedFilter[], data: Record<string, any>): SharedFilter[] => {
```
joshlacey, line 4: Can you add logic to this function to only run if the filter is a data filter? api filters handle parent relationships differently

joshlacey (changes requested):

packages/dashboard/src/helpers/updateChildFilters.ts
Atash3000: Added logic to filter out all data filters

packages/dashboard/src/helpers/updateChildFilters.ts
```diff
@@ -0,0 +1,48 @@
...
+      if (uniqChildValues.length > 0) {
+        updatedFilters[childIndex] = {
+          ...childFilter,
+          values: uniqChildValues
+        }
+      }
+    }
+  })
+
+  return [...updatedFilters, ...urlFilters]
```
joshlacey, line 47: Won't this reorder sharedFilter indexes? They should stay in the same place as the dashboardFilter visualizations are referencing those indexes.

Atash3000: @joshlacey  moat likely will re order and create mess. i will find solution for next sprint.

joshlacey: I think we need to find a solution before committing this code. Can we push this ticket to next sprint?

joshlacey:
@Atash3000 I think I found an alternative solution for this issue: change these lines in addValuesToFilters.ts

image

I'm working on this as part of a fix for DEV-10394

packages/dashboard/src/helpers/updateChildFilters.ts
adamdoe: @Atash3000 what is the status here?

Atash3000: @adamdoe  waitings for Josh to check it.

packages/dashboard/src/helpers/updateChildFilters.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (childFilterIndexes.length === 0) return newSharedFilters
+
+  // deep copy of the shared filters
+  const updatedFilters = _.cloneDeep(newSharedFilters)
+
+  // Update each child filter
+  childFilterIndexes.forEach(childIndex => {
+    const childFilter: SharedFilter = newSharedFilters[childIndex]
+    const parentFilter: SharedFilter = newSharedFilters.find(
+      filter => String(childFilter.parents) === String(filter.key)
```
joshlacey, line 20: childFilter.parents is an Array<string>

packages/dashboard/src/helpers/updateChildFilters.ts
```diff
@@ -0,0 +1,50 @@
...
+    )
+    const isParentMultiSelect = parentFilter.filterStyle === 'multi-select'
+
+    if (parentFilter) {
+      // Filter dataset based on parent's active value
+      const parentsActiveValues: string[] = dataSet.filter((d: Record<string, any>) => {
+        if (isParentMultiSelect) {
+          // If multi-select, check if the parent active array includes the value
+          return Array.isArray(parentFilter.active) && parentFilter.active.includes(d[parentFilter.columnName])
```
joshlacey, line 29: if isparentMultiSelect is trun the Array.isArray(parentFilter.active) should also always be true. but you could also simplify this by doing
`parentFilter.active?.includes(d[parentFilter.columnName])`
just incase it comes up undefined

packages/dashboard/src/helpers/updateChildFilters.ts
```diff
@@ -0,0 +1,50 @@
...
+        } else {
+          return parentFilter.active === d[parentFilter.columnName]
+        }
+      })
+      // Get unique active values for the child filter
+      const uniqChildValues = _.uniq(parentsActiveValues.map(d => d[childFilter.columnName]).filter(Boolean))
```
joshlacey, line 35: it would be helpful to name this just `const childFilterValues` I think the comment above is confusing actually because it looks like you mean that the child value itself is supposed to be active. when really you just mean the children of the active parents.

packages/dashboard/src/helpers/updateChildFilters.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      // Update the child filter if unique values exist
+      if (uniqChildValues.length > 0) {
+        const isChildMultiSelect = childFilter.filterStyle === 'multi-select'
+        const activeValue = isChildMultiSelect ? uniqChildValues : uniqChildValues[0]
```
joshlacey, line 40: do you mean to make the multiselect have all values active?

joshlacey (changes requested):

packages/dashboard/src/helpers/updateChildFilters.ts
Atash3000: I think it is strings. A child filter can not have multiple parents from what we have.

packages/dashboard/src/helpers/updateChildFilters.ts
Atash3000: it is not always array. and it was throwing error when i did parentFilter.active?.includes(d[parentFilter.columnName])

packages/dashboard/src/helpers/updateChildFilters.ts
Atash3000: true m in multi select we will show all values  childFilterValues .

packages/dashboard/src/helpers/tests/updatesChildFilters.test.ts
```diff
@@ -0,0 +1,56 @@
...
+        active: 'John',
+        key: 'Parent Filter',
+        values: ['John', 'Kelly', 'Norman', 'Jane']
+      },
+      {
+        tier: 2,
+        columnName: 'lastName',
+        active: '',
+        key: 'Child Filter',
+        parents: 'Parent Filter',
```
joshlacey, line 48: this should be an array

joshlacey (approved):

PR #1824, opened 2025-01-09
FIX:[DEV-2183] Charts Bottom Legend - Reverse Labels not working

Atash3000:
## Testing Steps
Open Chart 
On Legend panel  set Legend position to Top or Bottom
Click Single Row legend
Click Reverse Labels
Labels should revers back
2: Fixed Vertical sorting 
While legend is not Single row click Vertical Sort checkbox
it should sort legend items accordingly even if legend items(names)  are updated from data series panel
see attcached images 
Screenshot 2025-01-09 at 11 42 59
Screenshot 2025-01-09 at 11 43 17

## Self Review


packages/chart/src/components/Legend/helpers/createFormatLabels.tsx
```diff
@@ -93,31 +109,21 @@ export const createFormatLabels =
       return reverseLabels(seriesLabels)
     }
 
-    // DEV-4161: replaceable series name in the legend
```
adamdoe, line 96: Why was this removed?

packages/chart/src/scss/main.scss
```diff
@@ -50,13 +50,6 @@
   overflow-y: auto;
 }
 
-.d-flex {
```
adamdoe, line 53: Good find!

packages/chart/src/scss/main.scss
```diff
@@ -50,13 +50,6 @@
...
-  display: flex;
-}
-.flex-column-reverse {
```
adamdoe, line 56: Good!

adamdoe (commented):
@Atash3000 does series renaming still work in legends with that function removed?

packages/chart/src/components/Legend/helpers/createFormatLabels.tsx
Atash3000: replaced it with shorter verstion  if (config.series.some(item => item.name)) {
      const uniqueLabels = Array.from(new Set(config.series.map(d => d.name || d.dataKey))).map((val, i) => ({
        datum: val,
        index: i,
        text: val,
        value: colorScale(val)
      }))
      return reverseLabels(uniqueLabels)
    }

adamdoe (approved):

PR #1833, opened 2025-01-13
Feature:[DEV-9186a] Symbol line charts

Atash3000:
## Testing Steps
Added more Symbols to Line chart, Polygon and Cross
Screenshot 2025-01-13 at 15 19 41

## Self Review


packages/core/components/VisxShape.tsx
```diff
@@ -21,23 +17,41 @@ interface VisxShapeProps {
...
+  stroke,
+  key,
+  strokeWidth = 0
+}) => {
   if (!display) {
-    return null
+    return <g></g>
   }
 
   const Shape = glyphs[index]
```
adamdoe, line 41: I think instead of using an index here the shape name should be used.

packages/core/components/VisxShape.tsx
```diff
@@ -21,23 +17,41 @@ interface VisxShapeProps {
...
   const isFlipped = index === 4
+
   const transform = `translate(${left}, ${top}) ${isFlipped ? 'rotate(180)' : ''}`
 
   return (
     <g key={key} transform={transform}>
-      <Shape strokeWidth={0} fill={fill} stroke={stroke} size={size} />
+      {index === 6 ? (
+        <Text textAnchor='start' fill={fill} dy={3.5} dx={-5} fontSize={size / 4}>
```
adamdoe, line 49: This shape should be added as CustomGlyph from Visx. Then you can include it in the Glyphs array and remove the logic in this section:

'pentagon': CustomGlyph

packages/core/components/VisxShape.tsx
```diff
@@ -21,23 +17,41 @@ interface VisxShapeProps {
...
   display: boolean
-  index: GlyphKeys
+  index: number
   stroke: string
   key: string | number
+  strokeWidth?: number
 }
 
-const VisxShape: React.FC<VisxShapeProps> = ({ fill, size, left, top, display, index, stroke, key }) => {
+const VisxShape: React.FC<VisxShapeProps> = ({
```
adamdoe, line 26: Since visx handles these properties already it seems redundant to use a separate file here.

packages/chart/src/components/LineChart/components/LineChart.Shape.tsx
```diff
@@ -115,7 +115,7 @@ const LineChartShape = (props: LineChartShapeProps) => {
   if (config.lineDatapointStyle === 'always show') {
     return (
       <VisxShape
-        display={config.visual.lineDatapointSymbol === 'standard' && dataIndex !== 5}
+        display={config.visual.lineDatapointSymbol === 'standard' && dataIndex < 7}
```
adamdoe, line 118: Can we use a variable for the number seven that describes what seven is? Then we can export it to the other areas that use this number.

packages/chart/src/components/Legend/LegendLine.Shape.tsx
```diff
@@ -18,7 +18,7 @@ const LegendLineShape = props => {
 
       <VisxShape
         index={index}
-        display={config.legend.hasShape && index !== 5}
+        display={config.legend.hasShape && index < 7}
```
adamdoe, line 21: Use exported variable here as well.

packages/chart/src/components/EditorPanel/components/Panels/Panel.Visual.tsx
```diff
@@ -188,8 +188,8 @@ const PanelVisual: FC<PanelProps> = props => {
               updateField={updateField}
               options={['none', 'standard']}
             />
-            {config.series.length > 5 && config.visual.lineDatapointSymbol === 'standard' && (
-              <small className='text-danger'>Standard only supports up to 5 data points</small>
+            {config.series.length > 7 && config.visual.lineDatapointSymbol === 'standard' && (
+              <small className='text-danger'>Standard only supports up to 7 data points</small>
```
adamdoe, line 192: Use exported variable should it change.

packages/chart/examples/feature/tests-non-numerics/example-combo-bar-nonnumeric.json
```diff
@@ -4,44 +4,62 @@
   "theme": "theme-indigo",
```
adamdoe, line 4: Move to /src/stories/_mocks or similar folder.

adamdoe (changes requested):

packages/core/components/VisxShape.tsx
Atash3000: Because we use indexes of series key. using names would add more complexity.

packages/core/components/VisxShape.tsx
Atash3000: I am not sure if Visx supports custom glypth. how ?

packages/core/components/VisxShape.tsx
Atash3000: I removed file.

packages/core/components/VisxShape.tsx
adamdoe: I think you can follow the example in this file:
https://airbnb.io/visx/glyphs

The import is Glyph as CustomGlyph

adamdoe (approved):
Looks like you already found the Glyph Size change. I think you can merge this in now!

PR #1837, opened 2025-01-14
[DEV-9278] US maps with states, state text labels missing

khartco:
## [DEV-9278]

State labels enabled by default, add stroke color around state labels for increased contrast. Tooltips to recommend state labels be enabled.

## Testing Steps

- Create a new US map, or use a config without the `displayStateLabels` specifed.
  - State labels should be enabled by default
- Use a US map config with colored state background colors
  - On light colored states: State labels should appear dark gray with a white stroke 
  - On dark colored states: State labels should appear white with a gray stroke 
- In the editor, the toggle for state labels should have a new tooltip recommending to leave enabled.

## Self Review

## Screenshots (if applicable)

n/a

## Additional Notes

n/a


packages/core/helpers/cove/accessibility.ts
```diff
@@ -34,3 +34,19 @@ export const getColorContrast = (color1: string, color2: string) => {
   if (!chroma.valid(color1) || !chroma.valid(color2)) return false
   return chroma.contrast(color1, color2)
 }
+
+type OutlinedTextColors = {
+  textColor: string
+  strokeColor: string
+}
+
+export const getOutlinedContrastColors = (bgColor: string): OutlinedTextColors => {
```
adamdoe, line 43: I think this function might be doing the same thing as getColorContrast. For maintainability it might be easier to look at the two functions and figure out how to combine them. It's possible that the old function is using the harsh black color before the cool gray colors were introduced. Since one returns the original text color you would need to figure out a way to refactor where its used, or alternatively, return an object if an additional paramater is passed:

Currently returns a string
`export const getColorContrast = (color1: string, color2: string): string => {}`

Adding param to return the textColor/strokeColor combination:
`export const getColorContrast = (color1: string, color2: string, strokeColor: string): string | { textColor: string, strokeColor: string} => {}`

If you want to keep this function separate for now thats fine as well. 

nit: use const when variables don't change
nit: you can get variable color from the body if needed:

```.js
const bodyStyles = getComputedStyle(document.body)
const coolGray90 = bodyStyles.getPropertyValue('--cool-gray-90')
```

adamdoe (changes requested):

packages/core/helpers/cove/accessibility.ts
khartco: Going to keep the function separate for now, but I'll ask if other similar text should also be gray text. I guess a little nervous about regressions in those other areas if they're really intended to be full black or white.

packages/core/helpers/cove/accessibility.ts
```diff
@@ -34,3 +34,27 @@ export const getColorContrast = (color1: string, color2: string) => {
...
+  textColor: string
+  strokeColor: string
+}
+
+/**
+ * Get the ideal text and text-outline/stroke color for use on a given background color.
+ *
+ * @param {string} bgColor Color of the background
+ * @returns {OutlinedTextColors} Text and outline stroke color ideal for the background in hex (e.g. '#abc123') format
+ */
```
khartco, line 48: @adamdoe Do we have a preference of documentation style? I went with jsdoc here, but could keep it simpler too

adamdoe: I use jsdoc as well. We don't have defined rules for documentation but I find it useful when developers use some form of convention so this is great!  I think it comes down to the developers decision on whether to add additional docs - some are fine reading through the function and TS, but I like the additional hints/docs provided in the IDE. I think in TS projects you can leave off the types in JSDocs since TS is intelligent enough to infer them. This allows writing _somewhat_ cleaner descriptions. No need to change anything, just a fun nice-to-know thing.

`@param bgColor - color of the background`
`@returns text and outline stroke color...`

adamdoe (approved):

PR #1896, opened 2025-02-03
Enhancement [DEV 10400] Tab Simple

Mgetz10:
## [[DEV-10400](https://websupport-cdc.msappproxy.net/browse/DEV-10400)]

Create new filter style - Tab Simple based on the specs below

Screenshot 2025-02-03 at 3 23 16 PM

This filter should not convert to dropdown

## Testing Steps

[Tab simple story](http://localhost:6006/?path=/story/components-templates-map-filters--tab-simple)

## Self Review


packages/core/components/Filters/Filters.tsx
```diff
@@ -415,9 +422,10 @@ const Filters = (props: FilterProps) => {
         )
 
         Tabs.push(
-          <button
+          <h3
```
adamdoe, line 425: @Mgetz10 h3s can't have onClick handlers. 508 feedback will come back here if we pull this in.

adamdoe (changes requested):
The 508 concern is the only thing I'm seeing on this one. Nice work!

packages/core/components/Filters/Filters.tsx
Mgetz10: @adamdoe Maybe im misinterpreting Annettes comment (see description) any idea how I can accommodate while staying 508 compliant?

adamdoe: @Mgetz10 - it will need to be a button with h3 styling. For example, something like `<button className='h3' />`.

adamdoe (approved):

PR #1899, opened 2025-02-04
fix [dev-10241] shared filter keys repeating

joshlacey:
## Testing Steps

Scenario: upload [dev-10241.json](https://github.com/user-attachments/files/18662996/dev-10241.json) and navigate to dashboard preview.

1) change the race in the first "First" dropdown.
2) scroll down and change the second "First" dropdown to a different value that the first.

Expected: The charts should be filtered based on their respective filter value.

3) navigate to one of the dashboard filter widgets' editor panel.
4) add a two new filters and give them the same label.

Expected: The second filter you add should have the label you typed plus ' (1)'

5) add another filter with the same label.

Expected: The third filter you add should have the label you typed plus ' (2)'

## Self Review


Matt-DiPaolo (approved):
All testing steps worked properly

PR #1920, opened 2025-02-07
FIX : [DEV-10329] Brush Slider Malfunction for Long Years

Atash3000:
## Testing Steps
Fixed issues with brush slider. IT is very hard to test it since the issue only on wcms.

## Self Review

### DIFF
--- a/packages/chart/src/components/BrushChart.tsx
+++ b/packages/chart/src/components/BrushChart.tsx
@@ -112,8 +112,10 @@ const BrushChart = ({ xMax, yMax }: BrushChartProps) => {
       const endDate = _.get(_.last(finalData), config.xAxis.dataKey, '')
       // add custom blue colored handlers to each corners of brush
       svg.selectAll('.handle--custom').remove()
-      // append handler
-      const [formattedStartDate, formattedEndDate] = [startDate, endDate].map(date => formatDate(parseDate(date)))
+      // Parse and format the dates, setting them to an empty string if undefined
+      const parseAndFormatDate = date => (date ? formatDate(parseDate(date)) : '')
+      const formattedStartDate = parseAndFormatDate(startDate)
+      const formattedEndDate = parseAndFormatDate(endDate)
       svg.call(brushHandle, selection, formattedStartDate, formattedEndDate)
 
       setBrushConfig({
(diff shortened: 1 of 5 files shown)


packages/chart/src/components/Legend/helpers/index.ts
```diff
@@ -11,14 +11,19 @@ export const getGradientConfig = (config, formatLabels, colorScale) => {
 }
 
 export const getMarginTop = (isLegendBottom, config) => {
+  const DEFAULT_MARGIN_TOP = 27
```
adamdoe, line 14: We need better documentation when random numbers are used. Can you add a comment on why 27 is used here? Is it derived from margins/padding?

packages/chart/src/scss/main.scss
```diff
@@ -133,7 +133,7 @@
   .subtext--responsive-ticks,
   .section-subtext {
     &--brush-active {
-      margin-top: 3em !important;
+      margin-top: 5rem !important;
```
adamdoe, line 136: In the same place you're adding --brush-active can you use a bootstrap margin helper?

packages/chart/src/components/BrushChart.tsx
```diff
@@ -165,7 +168,19 @@ const BrushChart = ({ xMax, yMax }: BrushChartProps) => {
...
+      if (xMax > 0) {
+        initializeBrush()
+      } else {
+        //set a timeout to retry or handle the negative xMax scenario
+        setTimeout(checkAndInitializeBrush, 500)
+      }
+    }
+
+    checkAndInitializeBrush()
+  }, [xMax])
```
adamdoe, line 183: Add a cleanup function for setTimeout.

adamdoe (changes requested):
Added a couple comments on this one.

packages/chart/src/scss/main.scss
Atash3000: Bootsrap does does not support large margins. I tryied mt-10 or mt-8 did not work. only mt-4 supported but that is very low amount

adamdoe (changes requested):
Screen Shot 2025-02-12 at 12 15 26 PM
I'm seeing undefined on the brush slider and the brush slider overlapping into the data table area.

Atash3000:
@adamdoe  should be fixed undefined and margin issues.

PR #1926, opened 2025-02-12
[DEV-10402] Set Default Filter Value

adamdoe:
## [DEV-10402] Set Default Filter Value
For filters:

new field that is explicitly the default filter value that is separate from the "active" selection could potentially. This will allow users to select a default value without having to change the sort order.

## Testing Steps
On a shared filter there should be a new dropdown for setting the active value on data filters.

## Self Review


packages/dashboard/src/components/DashboardFilters/DashboardFiltersEditor/components/FilterEditor.tsx
```diff
@@ -433,6 +433,13 @@ const FilterEditor: React.FC<FilterEditorProps> = ({ filter, filterIndex, config
                     </select>
                   </label>
 
+                  <Select
+                    value={filter.active || '-Select-'}
```
joshlacey, line 437: no option to clear the default selection...

joshlacey (changes requested):

PR #1933, opened 2025-02-13
Feature: [DEV-10067] filter and filter button styles from tp5

superKazi:
## [[DEV-10067]](https://websupport.cdc.gov/browse/DEV-10067)

Updates filter dropdown select elements and buttons to match tp5 ux specs. filter options styles not affected.

## Testing Steps

- run `lerna run --scope @cdc/dashboard start`
- load `/examples/filtered-dash.json` config
- add ?editor=true to url and configure the dashboard so that filters require explicit apply and reset buttons
- preview updated dashboard and confirm css styles for fitler buttons and dropdowns are rendering as expected
- expected [dropdown styles](https://uxpreview.cdc.gov/84WBNF/?id=txxh1d&p=dropdowns)
- expected [filter button styles](https://uxpreview.cdc.gov/3GE5EW/?id=6cg4ol&p=fonts__spacing__and_elements_in_charts__ux-1249_&sc=3)

### DIFF
--- a/packages/core/components/Filters/Filters.tsx
+++ b/packages/core/components/Filters/Filters.tsx
@@ -218,7 +218,6 @@ export const useFilters = props => {
 
   const handleReset = e => {
     let newFilters = [...visualizationConfig.filters]
-    e.preventDefault()
 
     // reset to first item in values array.
     let needsQueryUpdate = false
@@ -249,8 +248,8 @@ export const useFilters = props => {
   }
 
   const filterConstants = {
-    buttonText: 'Apply Filters',
-    resetText: 'Reset All'
+    buttonText: 'Apply',
+    resetText: 'Clear Filters'
   }
 
   // prettier-ignore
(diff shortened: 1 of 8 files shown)


jayb (changes requested):
Looks like Josh has some refactoring ideas -- you guys can work those out. A few visual things I noticed:

* Dropdown `font-weight` is `400`, in the spec it's `300`

* There's some extra space between dropdown labels and the dropdown itself. Seems to be coming from the dropdown wrapper. Could we tighten it up to where it was? Relatedly: do we need the new dropdown wrapper? Maybe we do, but it might be nice to remove complexity if possible.

* Outside the scope of the ticket, but it could be nice if the buttons said "Clear Filter" instead of "Clear Filters" when there's only one filter.

* In the editor, some dropdown text is getting cut off (see the "g" in this screenshot). Haven't noticed it in an actual chart, but it's worth checking what's going on.

Screenshot 2025-02-14 at 9 25 32 AM

* In the attached config, text in the dropdown is overrunning the dropdown arrow. Possible to fix?

Screenshot 2025-02-14 at 9 38 22 AM

[trends-rsv.json](https://github.com/user-attachments/files/18800726/trends-rsv.json)

packages/core/components/Filters/Filters.tsx
```diff
@@ -218,7 +218,6 @@ export const useFilters = props => {
 
   const handleReset = e => {
     let newFilters = [...visualizationConfig.filters]
-    e.preventDefault()
```
jayb, line 221: Do we need to remove this?

PR #1968, opened 2025-02-27
Fix: Dev-10558 Toggle Data Viz disappears when toggle changes

Matt-DiPaolo:
## Dev-10558

The toggle rows do not disappear after switching between toggles

## Testing Steps

Scenario: Upload [oral-health-update.json](https://github.com/user-attachments/files/19012087/oral-health-update.json) and open Dashboard Preview
Expected: There are 4 dashboards and the current dashboard, All States, has 2 filters. 

1. Make a selection for both filters and Click Run Report 
Expected: The dashboard loads with a map in a toggle row. There are 3 buttons: Map, Chart, and Table 

2. Click the Table button
Expected: The Map will be replaced with a Table.

3. Click on the Dashboard tab Single County
Expected: The Single County Dashboard opens up and there are 5 filters

4. Make a selection for all 5 filters and click Run Report
Expected: The dashboard loads with a chart in a toggle row. There are 2 buttons: Table and Chart

5. Click the Table button
Expected: The Chart will be replaced with a Table.

## Self Review

## Additional Notes

When there were multiple dashboards and there was Toggles in each of them, the state from one was passed to the other causing toggles to break. Example: If you choose the 3rd option in a toggle row and then switched dashboards that has a toggle row with only 2 options, the toggle will still try to open the 3rd option even if there isn't a 3rd option. This caused the visualization to show a blank row.

### DIFF
--- a/packages/dashboard/src/components/VisualizationRow.tsx
+++ b/packages/dashboard/src/components/VisualizationRow.tsx
@@ -1,5 +1,5 @@
 import DataTableStandAlone from '@cdc/core/components/DataTable/DataTableStandAlone'
-import React, { useContext, useMemo } from 'react'
+import React, { useContext, useEffect, useMemo, useState } from 'react'
 import Toggle from './Toggle'
 import _ from 'lodash'
 import { ConfigRow } from '../types/ConfigRow'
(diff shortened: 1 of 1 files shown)


packages/dashboard/src/components/VisualizationRow.tsx
```diff
@@ -127,6 +136,9 @@ const VisualizationRow: React.FC<VizRowProps> = ({
   }
   return (
     <div className={`row${row.equalHeight ? ' equal-height' : ''}${row.toggle ? ' toggle' : ''}`} key={`row__${index}`}>
+      {row.toggle && !inNoDataState && (
+        <Toggle row={row} visualizations={config.visualizations} active={show.indexOf(true)} setToggled={setToggled} />
```
joshlacey, line 140: why would you just do active={toggledRow} ?

joshlacey (changes requested):

PR #1987, opened 2025-03-11
FIX:[DEV-50467] Remove bar Border from supressed & missing bars

Atash3000:
## Testing Steps
Open Vertical/Horizontal regular bar chart.
Update dataset value to "ABC" or zero( 0) to see missing bars.
In General Panel turn on checkboxes for zero value and missing data values to see bars that are missing
on visual panel turn on Bars Border
Result : Bars with missing values or zero values or suppressed values should not get any borders as shown on the below image:
Screenshot 2025-03-11 at 10 46 37

## Self Review

### DIFF
--- a/packages/chart/src/components/BarChart/components/BarChart.Vertical.tsx
+++ b/packages/chart/src/components/BarChart/components/BarChart.Vertical.tsx
@@ -144,7 +144,13 @@ export const BarChartVertical = () => {
                   <li class="tooltip-body ">${tooltipBody}</li>
                   ${additionalColTooltip ? '<li class="tooltip-body ">' + additionalColTooltip + '</li>' : ''}
                     </li></ul>`
-
+                  const { barHeight, isSuppressed, getBarY, getAbsentDataLabel } = getBarConfig({
+                    bar,
+                    defaultBarHeight,
+                    config,
+                    barWidth,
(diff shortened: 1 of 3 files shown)


packages/chart/src/components/BarChart/components/BarChart.Vertical.tsx
```diff
@@ -144,7 +144,13 @@ export const BarChartVertical = () => {
                   <li class="tooltip-body ">${tooltipBody}</li>
                   ${additionalColTooltip ? '<li class="tooltip-body ">' + additionalColTooltip + '</li>' : ''}
                     </li></ul>`
-
+                  const { barHeight, isSuppressed, getBarY, getAbsentDataLabel } = getBarConfig({
```
joshlacey, line 147: Why not return the absentDataLabel directly instead of returning a function?

Atash3000: Fixed!

joshlacey: seems like it's still returning a function?

Atash3000: @joshlacey  you probably dont have latest changes?

PR #2001, opened 2025-03-17
Feature:[DEV-10389] Update Legend Suppression Link Tooltip

Atash3000:
## Testing Steps
Open Bar Charts
Add suppression bars
Below Legend there is a link that notifies user about suppressed data
Hover on the link
after fixes : the link and tooltip information container was updated by size. font weight, shadow etc 
below is image: 
Screenshot 2025-03-17 at 18 44 14

## Self Review


adamdoe (changes requested):
Can we create a component called TooltipLink or something so that the behavior for this type of tooltip is repeatable?

Atash3000:
@adamdoe  created new component with it is css file in Core/TooltipLink. Let me know if it works.

packages/core/components/TooltipLink/TooltipLink.tsx
```diff
@@ -0,0 +1,39 @@
...
+                reported values does not meet the minimum
+                 reporting threshold.'
+        data-tooltip-id='supression-tooltip'
+        href={href}
+      >
+        {linkText}
+      </a>
+
+      <ReactTooltip
+        id='supression-tooltip'
```
adamdoe, line 22: I don't think you need these individual `<ReactTooltip />` components for each tooltip. You can likely reference the component thats found at the bottom of the Chart file. Right now I think thats ln 1538 in LinearChart:

```.js
          <ReactTooltip
            id={`cdc-open-viz-tooltip-${runtime.uniqueId}`}
            ...
```

adamdoe (approved):
After checking tooltips this looks good.

PR #2015, opened 2025-03-21
Fix [dev-10534] download image cutting

joshlacey:
## Testing Steps

upload:
[dev-10534-download-image.json](https://github.com/user-attachments/files/19391282/dev-10534-download-image.json) and navigate to dashboard preview.

1) click download image:

Expected: you should get an image download with none of the visualization cut off.

2) navigate to the chart editor and switch the legend from bottom to right.

3) click download image.

Expected: visualization shouldn't be cut off

## Self Review

### DIFF
--- a/packages/chart/src/components/AreaChart/components/AreaChart.jsx
+++ b/packages/chart/src/components/AreaChart/components/AreaChart.jsx
@@ -14,13 +14,28 @@ import { bisector } from 'd3-array'
 const AreaChart = props => {
   const { xScale, yScale, yMax, xMax, handleTooltipMouseOver, handleTooltipMouseOff, isDebug, children } = props
   // import data from context
-  let { transformedData, config, handleLineType, parseDate, formatDate, formatNumber, seriesHighlight, colorScale, rawData, brushConfig } = useContext(ConfigContext)
+  let {
+    transformedData,
+    config,
+    handleLineType,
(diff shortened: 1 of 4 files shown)


jayb (commented):
Is there a way to solve this problem without adding `10px` of padding to the top of every chart? Maybe it could be conditional on the user clicking the "download image" button?

We've worked hard to align our charts to recommendations from the UX team, and this deviates from their recommendations.

joshlacey:
@jayb the current logic allows the line chart to extend outside the bounds of the svg box itself. which is what is causing the image to get cut off as the query selector can't calculate the position of the inner items: 

image

I couldn't think of another solution than extending the bounding box of the svg itself. Can you think of another way?

jayb:
Looking through storybook and other configs, I'm not seeing other cases where elements overflow the top of the svg aside from the two examples you have here, so maybe there are specific chart elements that cause the overflow? A couple possible solutions are:

* Add the `10px` of padding when the user clicks the "download image" button
* Find another way to conditionally add the padding
* Fix the underlying issues that allow elements to overflow the svg. Maybe it's only CIs and curves that cause the problem?

Adding `10px` of space to the top of every COVE chart doesn't seem like a great solution to a pretty narrow bug (downloading a chart image that happens to overflow).

PR #2031, opened 2025-03-26
[DEV-50368] Maps > Reduce Use Effects

adamdoe:
### General Improvements
- Refactored multiple components and hooks to improve modularity, readability, and maintainability.
- New Constants in helpers/constants.ts for consistant viewbox and SVG calculations.
    - `SVG_WIDTH, SVG_HEIGHT, SVG_PADDING, SVG_VIEWBOX, HEADER_COLORS`
- New Hook: `useColumnNames` which extracts column names (geoColumnName, primaryColumnName, etc.) from the state for reuse across components.
    - `state.columns.geo.name as geoColumnName`
    - `state.columns.primary.name as primaryColumnName`
    - etc.
- General Destructuring improvements 
- General unused import/variable improvements
- Removes eslint ignore comments for better error handling in the future.

### Refactored addUIDs Helper:
- Ensures UIDs are added to rows based on geography type.
- Added checks to prevent errors when configObj.data is undefined.

### Refactored useMapLayers:
- Added TypeScript types for event handlers (MouseEvent, ChangeEvent).
- Improved error handling for fetching GeoJSON layers.

### Removes UseEffects
- Two useEffects in EditorPanel.ts didn't appear to be used
- useEffect in CityList.ts refactored to a variable
- Removes useEffects from each map type that handled re-centering the map. These don't appear to be needed.


packages/map/src/CdcMap.tsx
```diff
@@ -125,52 +125,10 @@ const CdcMap = ({
 
   // hooks
   const { currentViewport, dimensions, container, outerContainerRef } = useResizeObserver(isEditor)
-  const { handleSorting } = useFilters({ config: state, setConfig: setState })
   const { generateRuntimeLegend } = useGenerateRuntimeLegend(legendMemo, legendSpecialClassLastMemo)
   const { generateRuntimeFilters } = useGenerateRuntimeFilters(state)
   const { generateRuntimeData } = useGenerateRuntimeData(state)
 
-  // Use Effects
```
adamdoe, line 133: These use effects are now called directly in BubbleList > handleBubbleClick

packages/map/src/CdcMap.tsx
```diff
@@ -125,52 +125,10 @@ const CdcMap = ({
 
   // hooks
   const { currentViewport, dimensions, container, outerContainerRef } = useResizeObserver(isEditor)
-  const { handleSorting } = useFilters({ config: state, setConfig: setState })
```
adamdoe, line 128: Not used

packages/map/src/CdcMap.tsx
```diff
@@ -339,13 +297,6 @@ const CdcMap = ({
     }
   }, [state, container, runtimeData.init])
 
-  useEffect(() => {
```
adamdoe, line 342: Called directly in EditorPanel

packages/map/src/components/CityList.tsx
```diff
@@ -19,45 +20,42 @@ type CityListProps = {
...
   const { applyLegendToRow } = useApplyLegendToRow(legendMemo, legendSpecialClassLastMemo)
+  const { geoColumnName, latitudeColumnName, longitudeColumnName, primaryColumnName } = useColumnNames()
+  const { additionalCityStyles } = state.visual || []
+
   if (!projection) return
 
   const { geoClickHandler } = useGeoClickHandler()
   const { applyTooltipsToGeo } = useApplyTooltipsToGeo()
 
-  useEffect(() => {
```
adamdoe, line 30: Converted to variable for updating the citiesData

packages/map/src/components/EditorPanel/components/EditorPanel.tsx
```diff
@@ -71,20 +70,6 @@ const EditorPanel = () => {
   const [displayPanel, setDisplayPanel] = useState(true)
   const [activeFilterValueForDescription, setActiveFilterValueForDescription] = useState([0, 0])
 
-  const headerColors = [
```
adamdoe, line 74: Moved to constants

packages/map/src/components/EditorPanel/components/EditorPanel.tsx
```diff
@@ -1278,21 +1249,6 @@ const EditorPanel = () => {
     })
   }
 
-  useEffect(() => {
```
adamdoe, line 1281: Update the reversed color palettes directly below

packages/map/src/components/UsaMap/components/UsaMap.Region.tsx
```diff
@@ -64,16 +71,6 @@ const UsaRegionMap = props => {
     fetchData()
   }, [])
 
-  // When returning from another map we want to reset the state
```
adamdoe, line 67: Doesn't appear to be needed.

packages/map/src/components/WorldMap/WorldMap.tsx
```diff
@@ -57,13 +68,9 @@ const WorldMap = () => {
   // Resets to original data & zooms out
   const handleReset = (state, setState, setRuntimeData) => {
     const newRuntimeData = generateRuntimeData(state)
-    setRuntimeData(newRuntimeData)
-    setState({
-      ...state,
-      focusedCountry: false,
-      mapPosition: { coordinates: [0, 30], zoom: 1 }
-    })
+    setPosition({ coordinates: [0, 30], zoom: 1 })
```
adamdoe, line 71: setPosition is the correct zooming function here.

packages/map/src/components/EditorPanel/components/EditorPanel.tsx
```diff
@@ -1247,8 +1220,6 @@ const EditorPanel = () => {
     })
   }
 
-  const usedFilterColumns = {}
```
adamdoe, line 1250: Not used

packages/map/src/components/EditorPanel/components/EditorPanel.tsx
```diff
@@ -1278,21 +1249,6 @@ const EditorPanel = () => {
     })
   }
 
-  useEffect(() => {
-    const parsedData = convertStateToConfig()
-    const formattedData = JSON.stringify(parsedData, undefined, 2)
```
adamdoe, line 1283: Not used

packages/map/src/CdcMap.tsx
```diff
@@ -635,34 +587,33 @@ const CdcMap = ({
                 general.type !== 'navigation' &&
                 false === loading && (
                   <DataTable
-                    config={state}
```
adamdoe, line 638: Sorting

packages/map/src/hooks/useGenerateRuntimeData.ts
```diff
@@ -5,7 +5,23 @@ import { VizFilter } from '@cdc/core/types/VizFilter'
 import { DataRow } from '../types/MapConfig'
 
 const useGenerateRuntimeData = (state: MapConfig) => {
-  const generateRuntimeData = (configObj: MapConfig, filters: VizFilter[], hash: number, test?: boolean) => {
+  /**
+   * Generates runtime data based on the provided configuration object, filters, and hash.
+   *
+   * @param {MapConfig} configObj - The configuration object for the map.
```
joshlacey, line 11: I wouldn't add all these comments for each param because they could change in the future and you'd need to update them in two places. I think the typescript itself is self documenting enough here.

packages/map/src/hooks/useColumnNames.ts
```diff
@@ -0,0 +1,14 @@
+import { useContext } from 'react'
+import ConfigContext from '../context'
+
+const useColumnNames = () => {
```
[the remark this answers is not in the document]
adamdoe: I think since we have a low number of columns, that don't change frequently, using a hook might be the better approach. We would have to set up boiler plate for passing in state every time as well. Probably a chicken and egg approach but I think having 5-6 named columsn returned from the hook might be better than having this type of boiler plate for every getter:

const geoColumnName = getColumnName('geo', state)

We would also run into situations where the variable would be named differently:

const geographyColumnName = getColumnName('geo', state)

I think because of this continuing to use the hook might be the better approach. If we need to dynamically fetch column names in the future I can see where having a getter would be beneficial. Do you have a strong opinion now based on what I've shared?

packages/map/index.html
```diff
@@ -20,7 +20,7 @@
     <!-- DEFAULT EXAMPLES -->
     <!-- <div class="react-container" data-config="/examples/private/nhis.json"></div> -->
```
joshlacey, line 21: remove this file from your git history

joshlacey (changes requested):

packages/map/src/hooks/useColumnNames.ts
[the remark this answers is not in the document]
joshlacey: I was suggesting this as a helper function: 

const getColumnNames = (columns: Columns): Record<string, string> => {
  const geoColumnName = columns.geo.name || null
  const primaryColumnName = columns.primary.name || null
  const latitudeColumnName = columns.latitude.name || null
  const longitudeColumnName = columns.longitude.name || null
  const categoricalColumnName = columns.categorical?.name || null
  return { geoColumnName, primaryColumnName, latitudeColumnName, longitudeColumnName, categoricalColumnName }

}

no need to tie this into context it seems like it's a simple getter function.

packages/map/src/hooks/useMapLayers.tsx
```diff
@@ -25,48 +30,43 @@ export default function useMapLayers(config: MapConfig, setConfig, pathGenerator
   const [featureArray, setFeatureArray] = useState([])
 
   useEffect(() => {
-    fetchGeoJSONLayers()
-  }, []) //eslint-disable-line
+    const fetchLayers = async () => {
+      try {
+        await fetchGeoJSONLayers()
+      } catch (error) {
```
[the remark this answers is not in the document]
joshlacey: did you mean to resolve this?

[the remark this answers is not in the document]
adamdoe: I think you might have commented on an outdated file.

packages/map/index.html
adamdoe: Reset file to dev here:
https://github.com/CDCgov/cdc-open-viz/pull/2031/commits/35f88f470e235c4988b35116dd078b0dfdb72dd0

packages/map/src/hooks/useMapLayers.tsx
```diff
@@ -25,48 +30,43 @@ export default function useMapLayers(config: MapConfig, setConfig, pathGenerator
   const [featureArray, setFeatureArray] = useState([])
 
   useEffect(() => {
-    fetchGeoJSONLayers()
-  }, []) //eslint-disable-line
+    const fetchLayers = async () => {
+      await fetchGeoJSONLayers()
```
joshlacey, line 34: can't you just call fetchGeoJSONLayers() now? you aren't awaiting for anything

joshlacey (approved):
looks good, was able to run in storybook and via the package

PR #2039, opened 2025-03-27
FIX: [DEV-10387] - Fix Widget delete issue

Atash3000:
## Testing Steps
Open Dashboards. 
When user deletes last widget in a row (clicking inner X button)  it should delete whole row.
Before: it used to delete width but row was left.
See Image: 
Screenshot 2025-03-27 at 15 00 00

## Self Review


packages/dashboard/src/components/Widget/Widget.tsx
```diff
@@ -71,7 +72,24 @@ const Widget = ({
...
   const deleteWidget = () => {
     if (!widgetConfig) return
+    // if last widget left in row remove whole row on delete
+    if (widgetConfig.colIdx === 0) {
+      let newVisualizations = { ...config.visualizations }
+      const rows = config.rows
+      const rowIdx = widgetConfig.rowIdx
+
+      //delete the instantiated widgets
+      if (rows[rowIdx] && rows[rowIdx].columns && rows[rowIdx].columns.length && config.visualizations) {
```
joshlacey, line 82: can simplify this: rows[rowIdx]?.columns?.length

packages/dashboard/src/components/Widget/Widget.tsx
```diff
@@ -71,7 +72,24 @@ const Widget = ({
 
   const deleteWidget = () => {
     if (!widgetConfig) return
+    // if last widget left in row remove whole row on delete
+    if (widgetConfig.colIdx === 0) {
```
joshlacey, line 76: what if it's the last column but it's at index 2?

packages/dashboard/src/components/Widget/Widget.tsx
```diff
@@ -71,7 +72,24 @@ const Widget = ({
...
+      //delete the instantiated widgets
+      if (rows[rowIdx] && rows[rowIdx].columns && rows[rowIdx].columns.length && config.visualizations) {
+        rows[rowIdx].columns.forEach(column => {
+          if (column.widget) {
+            delete newVisualizations[column.widget]
+          }
+        })
+      }
 
+      config.rows.splice(rowIdx, 1) // delete the row
```
joshlacey, line 90: I think the functionality should be that you delete the column but leave a placeholder there, a single column so that you can still add a widget there if you want. 

say for example you put a chart widget when you meant to put a table widget, it would be annoying if the row disappeared when you deleted the mistaken chart widget. you'd have to add a new row and if you have a lot of rows, move it all the way to the top.

joshlacey (changes requested):

packages/dashboard/src/components/Widget/Widget.tsx
Atash3000: @joshlacey  thats not issue. it will delete only last(one) widget.

packages/dashboard/src/components/Widget/Widget.tsx
```diff
@@ -71,7 +72,24 @@ const Widget = ({
...
+      updateConfig({ ...config, rows, visualizations: newVisualizations })
```
joshlacey, line 91: Can you match the pattern everywhere else here and call the dispatch function directly?

adamdoe:
@Atash3000 what is the status of this PR?

PR #2049, opened 2025-04-01
RE-FIX: [DEV-10142a] add right side padding to spark-line component

Atash3000:
## Testing Steps
Added padding to the right side of Spark Line : 
Screenshot 2025-04-01 at 15 48 13

## Self Review


adamdoe (changes requested):
[small fix] The padding should be applied to both sides using D3/Visx functions inside of the container or applied to both side around the container with CSS equally. This appears to have unbalanced padding.

Atash3000:
@adamdoe  update margin from inside and removing padding on outer container   const margin = { top: 5, right: 20, bottom: 10, left: 10 }

PR #2055, opened 2025-04-03
[pr-template] Refine pr template and workflow

adamdoe:
## Self Review


adamdoe:
> @adamdoe do you want the content of what's in the ticket to be copy pasted into the summary?

@joshlacey I think the goal is to require some text. Copy/Pasting works, or at least a brief description.

.github/pull_request_template.md
```diff
@@ -1,23 +1,27 @@
...
 
-<!-- Provide a brief description of the changes made in this PR -->
+## [NEW-XXXX] Summary
+<!-- Provide a brief explanation of the changes -->
 
-## Testing Steps
 
-<!-- Provide testing steps and reference storybook stories if necessary -->
-<!-- Add applicable configs to JIRA ticket for testers-->
+## How did you test this change?
```
joshlacey, line 10: I like Testing Steps better. I want to get easy instructions for booting up the change and seeing the change rather than some explanation.

.github/workflows/pr_template_validation.yml
```diff
@@ -0,0 +1,109 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v3
+
+      - name: Ensure Ticket Number is in Summary Section
+        run: |
+          PR_BODY=$(curl -s https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }} | jq -r '.body')
+
+          if ! echo "$PR_BODY" | grep -Pzo "(?s)^## (\[NEW-\d+\] Summary|\[NO-TICKET\] Summary|Summary)"; then
```
joshlacey, line 18: I think NEW- prefix is only going to be for while we're migrating to the new ticketing system maybe replace this with `\w{3,5}` to match 3-5 word characters make it more generic / future proof

joshlacey (changes requested):

.github/pull_request_template.md
adamdoe: Agreed - removed required checks on the summary section.

.github/workflows/pr_template_validation.yml
adamdoe: Added up to seven characters allowed on the prefix, that would support [FEATURE-123]. I also removed case sensitivity, so if folks use [feature-123], [BUG-123], or just summary in the line they shouldn't have any issues.

.github/pull_request_template.md
joshlacey: no like I want this to say "Testing Steps" again

adamdoe: Replaced the text with "Testing Steps" and updated the job that checks for that text.

joshlacey (approved):

PR #2063, opened 2025-04-09
Feature: [dev-9909] Filter Footnotes using Viz Filters

joshlacey:
Filter footnotes using Viz Filters

## Testing Steps

###Test 1
1) Select Dashboard Visualization.
2) Select a Data set.
3) select toggle at the top of the row. 
4) drag and drop a table and a chart onto the toggle row and map the selected dataset.
5) add visualization filters to both table and chart and select "filter footnotes"
6) configure footnotes by selecting the same datasource. Map the footnotes symbol and text to some columns in the dataset.
7) Navigate to Dashboard Preview.

Expected: Selecting a filter option should change the footnotes that are visible. Toggling between the table and the chart should show different footnotes depending on the respective selection on the visualization filter.

###Test 2 (Regression)

1) Load:
[states-and-territories-map.json](https://github.com/user-attachments/files/19869995/states-and-territories-map.json)
2) scroll all the way to the bottom of the page.

Expected: there should be a single visualization row of footnotes (there will be two separate footnotes in the row). under three separate Markup includes. (The footnotes on the row were moved to a new markup include row with a footnotes section via a Migration).


adamdoe (approved):
This is working for viz level filters. I did notice that shared filters don't seem to work but that could be outside of scope for this ticket.

PR #2076, opened 2025-04-21
Feature : [DEV-50357] - Add Grouped Legend to the Maps

Atash3000:
## Testing Steps
Open Maps
On the legend panel there is new dropdown "Legend Group By"
Select any value which groups legend by year, week, value etc
Note : Legend grouping works only if it is Categorical Maps
Also created a story, see attcahed image : 
Screenshot 2025-04-21 at 09 42 58


packages/map/src/components/Legend/components/LegendGroup/Legend.Group.css
adamdoe, line 1: [fix]: rename and fix import to the lowercase version based on new convention of having styles start lowercased.
legend.group.css

packages/map/src/components/Legend/components/LegendGroup/Legend.Group.tsx
```diff
@@ -0,0 +1,130 @@
+import { useContext, useMemo } from 'react'
+import './Legend.Group.css'
```
adamdoe, line 2: [fix]: this will need to be updated to the new name.

packages/map/src/components/Legend/components/LegendGroup/Legend.Group.tsx
```diff
@@ -0,0 +1,130 @@
...
+}
+
+interface GroupedData {
+  [key: string]: LegendItem[]
+}
+
+const LegendGroup = ({ legendItems }) => {
+  const { runtimeLegend, setAccessibleStatus, setRuntimeLegend, state, currentViewport } = useContext(ConfigContext)
+
+  const getGridColClass = (viewport: string) => {
```
adamdoe, line 22: Instead of using the viewport can you use bootstraps helpers?

col-number, col-md-number, col-lg-number?

adamdoe (changes requested):

PR #2077, opened 2025-04-21
Fix/ [NEW-50398] tooltips horizontal bar charts

joshlacey:
## Summary

Tool tips were not working on horizontal bar charts which were using a toggle in dashboard on the initial render. I tracked that down to an issue with two separate mouse event handlers in different hooks. I combined the handlers in the useBarChart hook so that they are both called together. 

## Testing Steps

upload: 
[new-50398-tooltip.json](https://github.com/user-attachments/files/19851589/new-50398-tooltip.json)

1) navigate to preview page.
2) toggle the bar chart.
3) hover over one of the bars.

Expected: you should see a tooltip with corresponding information.


adamdoe (approved):
Looks good!

PR #2093, opened 2025-05-05
Feature [NEW 50401] Standalone data table

Mgetz10:
## Summary
Standalone data table

## Testing Steps
`lerna run --scope @cdc/data-table start`


packages/data-table/src/helpers/fetchers.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const fetchConfig = (configUrl: string) =>
+  fetch(configUrl)
+    .then(res => res.json())
+    .catch(err => {
+      console.error('Error fetching config:', err)
+    })
+
+export const fetchData = (dataUrl: string) =>
+  fetchRemoteData(`${dataUrl}?${cacheBustingString()}`).catch(err => {
```
joshlacey, line 12: is the cachebustinging needed?

packages/data-table/src/helpers/fetchers.ts
```diff
@@ -0,0 +1,14 @@
+import cacheBustingString from '@cdc/core/helpers/cacheBustingString'
+import fetchRemoteData from '@cdc/core/helpers/fetchRemoteData'
+
```
joshlacey, line 3: I think these helpers might be a little too much abstraction. I see here that there's catch statements on fetchConfig for example but then I see that you're trying to also catch in the CdcDataTable file itself where they are called.

in CdcDashboard for example we are just doing
`const config = (await (await fetch(configUrl)).json())` I'm not sure what the error state would look like for that, but I haven't heard any complaints yet?

packages/data-table/src/CdcDataTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+    fetchData(dataUrl)
+      .then(resData => {
+        const processedData = processData(resData, dataDescription)
+        setData(processedData)
+      })
+      .catch(() => setData(null))
+  }, [inputData, dataUrl, configLoading, invalidConfig, dataDescription])
+
+  /* HANDLE LOADING/ERROR STATES */
+  if (configLoading || dataLoading) return <Loading />
```
joshlacey, line 124: there's should just be one loading state.

packages/data-table/src/CdcDataTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+  if (invalidConfig || invalidData) return <div>Something went wrong</div>
```
joshlacey, line 125: I think the current behavior would just be a blank screen. for consistency I would say maybe we want to do the same here and create a separate enhancement that will handle the error messages consistently across COVE?

packages/data-table/src/CdcDataTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  // initial static config combined with dynamic states
+  const configWithStates = {
```
joshlacey, line 128: this should be `state` and come from the reducer.

packages/data-table/src/CdcDataTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+        />
+      )}
+
+      {/* FILTERS */}
+      <div className='bg-white z-1'>
+        {filters && (
+          <Filters
+            config={configWithStates as unknown as Visualization}
+            setConfig={updateFilters}
+            setFilteredData={() => {}}
```
joshlacey, line 164: is this intentional?

packages/data-table/src/CdcDataTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+  return (
+    <Layout.VisualizationWrapper
+      ref={outerContainerRef}
+      config={config}
+      isEditor={isEditor}
+      showEditorPanel={showEditorPanel}
+      currentViewport={currentViewport}
+    >
+      {/* EDITOR */}
+      {isEditor && (
```
joshlacey, line 146: this section with EditorPanel Filters and DataTable seems really close to DataTableStandAlone could that have been used?

packages/data-table/src/components/EditorPanel.jsx
```diff
@@ -0,0 +1,129 @@
...
+  AccordionItemPanel
+} from 'react-accessible-accordion'
+
+import ColumnsEditor from '@cdc/core/components/EditorPanel/ColumnsEditor'
+import DataTableEditor from '@cdc/core/components/EditorPanel/DataTableEditor'
+import VizFilterEditor from '@cdc/core/components/EditorPanel/VizFilterEditor'
+import ErrorBoundary from '@cdc/core/components/ErrorBoundary'
+import Layout from '@cdc/core/components/Layout'
+
+const EditorPanel = ({
```
joshlacey, line 17: can you not reuse DataTableEditorPanel?

packages/data-table/src/index.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
```
joshlacey, line 1: .tsx

packages/data-table/src/test/CdcDataTable.test.jsx
```diff
@@ -0,0 +1,6 @@
+// Placeholder test until we add them in.
```
joshlacey, line 1: Can we add a test or remove this?

packages/data-table/src/types/Config.ts
```diff
@@ -0,0 +1,28 @@
+import { Column } from '@cdc/core/types/Column'
+import { DataDescription } from '@cdc/core/types/DataDescription'
+import { General } from '@cdc/core/types/General'
+import { Table } from '@cdc/core/types/Table'
+import { Version } from '@cdc/core/types/Version'
+import { VizFilter } from '@cdc/core/types/VizFilter'
+
+export type Config = {
```
joshlacey, line 8: core/components/DataTable/types/TableConfig.ts might be a duplicate of this.

packages/editor/src/components/ConfigureTab.tsx
```diff
@@ -55,6 +56,12 @@ export default function ConfigureTab({ containerEl }) {
           <CdcMarkupInclude isEditor={true} isDebug={isDebug} config={config} setConfig={setTempConfig} />
         </ErrorBoundary>
       )
+    case 'table':
+      return (
+        <ErrorBoundary component='CdcChart'>
```
joshlacey, line 61: wrong Component

packages/data-table/src/_stories/DataTable.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+const meta: Meta<typeof CdcDataTable> = {
+  title: 'Components/Templates/DataTable',
+  component: CdcDataTable
+}
+
+type Story = StoryObj<typeof CdcDataTable>
+
+export const DataTable: Story = {
+  args: {
+    config: editConfigKeys(DataTableConfig, [{ path: ['filters'], value: [] }]),
```
joshlacey, line 16: this is much more readable to me:

{ ...DataTableConfig, filters: [] }

is that not what you're trying to do here?

packages/data-table/src/helpers/configHelpers.ts
```diff
@@ -0,0 +1,28 @@
+import { cloneDeep } from 'lodash'
+import { ChartConfig } from '../types/ChartConfig'
+
+/* editConfigKeys
```
joshlacey, line 4: I feel like it's much easier to use Object.assign or the spread operator.

{....config, filters: []}

{...config, table: {...config.table, show: true}

if you really need to set something deeply nested lodash already has this function:

_.set(obj, ['x', 'y', 'z'], 'hello');

https://lodash.com/docs/4.17.15#set

joshlacey (changes requested):

packages/data-table/src/CdcDataTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+  // Load data when component first mounts
+  const outerContainerRef = useCallback(node => {
+    if (node === null) return
+    resizeObserver.observe(node)
+  }, [])
+
+  useEffect(() => {
+    /* NO VALID INPUT */
+    if (!inputConfig && !configUrl) {
+      setConfig(null)
```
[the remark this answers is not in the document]
Mgetz10: I'm using it to determine some vars as well 
` /* LOADING VARS */
  const configLoading = configUrl && config === undefined
  const dataLoading = data === undefined
  /* INVALID VARS */
  const invalidConfig = config === null
  const invalidData = data === null`

packages/data-table/src/components/EditorPanel.jsx
Mgetz10: I was hoping to get away from the master config workflow where the 'config' represents the state and instead have the config just represent the initial arguments and create state items for the parts that are dynamic. The reasoning behind this move was to make the state updating actions be more explicit and less of a black box. config is so big and is updated in so many places it can make tracing logic pretty difficult at times. in an ideal world the more explicit approach would also create less state updates but ultimately that would require more parts of the app to conform to this model. Anyway all this to say: thats why some of these parts look different - but in going through the feedback, which was understandably mostly about conforming to the way cove already works, the logic is becoming more redundant so im going to try plugging in DataTableStandAlone after-all.

packages/data-table/src/types/Config.ts
Mgetz10: This is a very specific config as trimmed down as i can get it that exactly spells out any keys that could be used in a standalone table viz.

packages/data-table/src/CdcDataTable.tsx
```diff
@@ -0,0 +1,196 @@
...
+  configUrl?: string
+  isEditor: boolean
+}
+
+const CdcDataTable = ({ config: inputConfig, configUrl, isEditor }: CdcDataTableProps) => {
+  /* STORE */
+  const initialState = getInitialState(isEditor)
+  const [state, dispatch] = useReducer(reducer, initialState)
+
+  const builtSetter = (type: ActionType) => payload => {
```
joshlacey, line 32: this function is not necessary. dispatch is already bound to the specific action types that you defined in the dataTable.actions.ts folder. Your IntelliSense should be giving options of types to choose from as you're calling dispatch.

joshlacey (changes requested):

packages/data-table/src/CdcDataTable.tsx
Mgetz10: not sure i understand the issue here - this is just a qol function to reduce repetitiveness.   

buildSetter('SET_CONFIG')
instead of
payload => dispatch({ type: 'SET_CONFIG', payload })

joshlacey: here's some more reasons and then I'll leave it up to you:

1) If you need to call dispatch from a deeply nested component these alias functions probably wouldn't be used: https://github.com/CDCgov/cdc-open-viz/blob/main/packages/dashboard/src/components/Widget/Widget.tsx What's great about dispatch context is it can be pulled in anywhere underneath this Component tree. Legacy code had a lot of state change functions that were prop drilled really deeply down the tree, dispatch was introduced to fix that prop drilling.

2) buildSetter factory function is a layer of abstraction that doesn't exist anywhere else in the code. 

3) There are other places in the app where dispatch alias functions were created because existing calls to those alias functions existed before the reducer pattern was introduced, imo I think even those in most cases should have just called dispatch directly.

PR #2099, opened 2025-05-07
Enhancement: [NEW-50418] Dashboard Dataset Override

joshlacey:
## Summary

This feature basically allows and external application to hijack a dataset on the config and reset the data. below you'll see `{'local_test.csv': newData}` this looks for the dataset with the key of local_test.csv and resets its data in dashboard. 

**Note there is no need to update the UI here. I have included in the PR 2 commits. the first is the above described feature. The second is some refactor work I did while trying to work on the sessionStorage option, which would have required a UI update.

## Testing Steps

Scenario: load the standalone dashboard package with 
[local_test.json](https://github.com/user-attachments/files/20093506/local_test.json)

in browser console paste the following: 

```
const newData = [{category: 'a', data: 1}]
const customEvent = new CustomEvent('cove_set_data', {detail: {'local_test.csv': newData}})
document.dispatchEvent(customEvent)
```

Expected: The table should reduce to just one row of data.


jayb (commented):
This looks good, the data updating piece is nice and simple.

Have you considered what happens if there are multiple dashboards on the page? You might want to 1) check that the dataset already exists in that dashboard before updating it, or 2) have a way to target the dashboard you're trying to set data for by adding something to the message that's unique to that dashboard.

joshlacey:
@jayb  I will add your suggestion #1 I like that. I think it could be used as a solution for #2, the developer can just namespace their data sources by which dashboard is using it. Any further check than that I think might be out of scope of this ticket.

PR #2106, opened 2025-05-09
[DEV-8148] fix: equal number opt in button

adamdoe:
## Summary
When legends are set to equal number the numbers shouldn't repeat, ie [10-11, 11-12, 12-13]. Instead they should start at the next number [10-11, 12-13, 14-15]

## Testing Steps
- Open story: http://localhost:6006/?path=/story/components-templates-map--equal-number-map
- The legend items should repeat [12-40, 40-55]
- Under legend type > change the legend to equal interval, then back to quantile. They shouldn't repeat now.


Atash3000 (approved):

Atash3000:
@adamdoe  Works properly.

PR #2111, opened 2025-05-12
Fix/map refactor 2

adamdoe:

### DIFF
--- a/packages/map/examples/color-fix.json
+++ b/packages/map/examples/color-fix.json
@@ -1,0 +1,761 @@
+{
+  "general": {
+    "geoType": "us",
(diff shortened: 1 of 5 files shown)


packages/map/index.html
```diff
@@ -30,7 +30,7 @@
```
joshlacey, line 30: remove from git history

packages/map/examples/color-fix.json
```diff
@@ -0,0 +1,761 @@
+{
```
joshlacey, line 1: remove from git history

joshlacey (changes requested):
it looks like this is working now. can you remove the examples jsons?

PR #2139, opened 2025-05-21
Fix/map refactor josh

joshlacey:


packages/map/src/components/WorldMap/WorldMap.tsx
```diff
@@ -191,12 +190,7 @@ const WorldMap = () => {
     <ErrorBoundary component='WorldMap'>
       {allowMapZoom ? (
         <svg viewBox={SVG_VIEWBOX} role='img' aria-label={handleMapAriaLabels(config)}>
-          <rect
-            height={SVG_HEIGHT}
-            width={SVG_WIDTH}
-            onClick={() => handleReset(config, setRuntimeData)}
-            fill='white'
-          />
+          <rect height={SVG_HEIGHT} width={SVG_WIDTH} onClick={() => handleReset(setRuntimeData)} fill='white' />
```
joshlacey, line 193: While we're here can we refactor this so that we aren't currying the setRuntimeData function for no reason?

PR #2145, opened 2025-05-28
Fix [new-50455] default multiselection

joshlacey:
## Summary

Changed multiselect to always provide the default selection once all other options are removed.

## Testing Steps

Scenario: run storybook and navigate to Pages > Dashboard > Pivot Filter.

1) select hispanic and black.
2) remove hispanic
3) remove black

Expected: the multiselect should default back to hispanic


packages/dashboard/src/components/DashboardFilters/DashboardFilters.tsx
```diff
@@ -1,205 +1,209 @@
...
-        >
-          {applyFiltersButtonText || 'GO!'}
-        </button>
-      )}
-    </form>
-  )
-}
-
-export default DashboardFilters
+import React from 'react'
```
joshlacey, line 1: for some reason the precommit hook is changing the line endings in this file. I think maybe at some point a PR went in that mixed the line endings. and the precommit hook is changing it to crlf like it should be?

packages/dashboard/src/components/DashboardFilters/DashboardFilters.tsx
```diff
@@ -1,205 +1,209 @@
...
+            {filter.filterStyle === FILTER_STYLE.multiSelect ? (
+              <MultiSelect
+                label={label}
+                options={multiValues}
+                fieldName={filterIndex}
+                updateField={(_section, _subsection, fieldName, val) => {
+                  const defaultSelection = [multiValues[0].value]
+                  const selection = val?.length ? val : defaultSelection
+                  updateField(null, null, fieldName, selection)
+                }}
```
joshlacey, line 149: the only changes made to this file are lines 146-148. All other changes are line endings.

adamdoe:
@joshlacey when I open this example and change the value to a new default value its not returning the correct value when all options are removed. For example:

- change the default to `Non-Hispanic American Indian` 
- add all values
- remove all values
- `Hispanic or Latino` shows as the default value

packages/dashboard/src/components/DashboardFilters/DashboardFilters.tsx
joshlacey: multiValues is an array constant. can't be reassigned to null

PR #2147, opened 2025-05-29
Feature/dev 8148b

adamdoe:
## Summary
Set config.general.equalNumberOptIn to true by default for all maps.
Note: This may affect the overall color and appearance of the maps, but the change has been reviewed and approved as acceptable moving forward.

## Testing Steps
Review this config locally: https://www.cdc.gov/oral-health-data-systems/oral_health_data/oral-health-dashboard.html
Choose All Counties Tab > then Intervention Strategies, Percent of PWS Population, and 2022. The legend items should match.

### DIFF
--- a/packages/map/src/helpers/generateRuntimeLegendHash.ts
+++ b/packages/map/src/helpers/generateRuntimeLegendHash.ts
@@ -4,7 +4,7 @@ import { MapConfig } from '../types/MapConfig'
 export const generateRuntimeLegendHash = (config: MapConfig, runtimeFilters) => {
   return hashObj({
     unified: config.legend.unified ?? false,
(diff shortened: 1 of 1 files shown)


Atash3000:
@adamdoe  on this  function generateRuntimeLegendHash has equalNumberOptIn: config.general.equalNumberOptIn ?? false. if general.equalNumberOptIn  by default is true then should we also update false to true?

adamdoe:
> @adamdoe on this function generateRuntimeLegendHash has equalNumberOptIn: config.general.equalNumberOptIn ?? false. if general.equalNumberOptIn by default is true then should we also update false to true?

Good catch, updated!

PR #2148, opened 2025-05-29
Feature/fix legend row

Atash3000:
## Testing Steps
Open maps
On the legend panel click 'single column" or single row legend depeing if  legend on the side or top/bottom
the checkbox was not working. after fix : checkbox is clickable.


adamdoe (changes requested):
fix: on this branch the defaults aren't aligned in two columns when the example loads.

PR #2171, opened 2025-06-11
FIX [DEV-50422] - update scale for Area chart

Atash3000:
## Testing Steps
Open Area chart.
Switch to 'Date scale"
Before there is padding on both sides of the chart
After : no padding, chart is fully drawn on xMax
See images : 
Screenshot 2025-06-11 at 16 59 14
Screenshot 2025-06-11 at 16 59 26


packages/chart/src/hooks/useScales.ts
```diff
@@ -77,6 +77,15 @@ const useScales = (properties: useScaleProps) => {
   if (config.xAxis.type === 'date' && !isHorizontal) {
     const xAxisDataMappedSorted = sortXAxisData(xAxisDataMapped, config.xAxis.sortByRecentDate)
     xScale = composeScaleBand(xAxisDataMappedSorted, [0, xMax], 1 - config.barThickness)
+    if (visualizationType === 'Area Chart') {
+      xScale = scalePoint({
+        domain: xAxisDataMappedSorted,
+        range: [0, xMax],
+        padding: 0
+      })
+      xScale.type = scaleTypes.POINT
```
jayb, line 86: Won't this always get overwritten by line 88?

Atash3000: Fixed.

PR #2182, opened 2025-06-17
FIX [DEV-10351] - (Brush Chart) Apply Button Jumps Chart off Axis

Atash3000:
## Testing Steps
Created new Brush chart using VISX library.
Updated visx/brush package to latest version.
Upload file and see if brush works 
Here is file : 
[DEV-10996.json](https://github.com/user-attachments/files/20783411/DEV-10996.json)

### DIFF
--- a/packages/chart/src/store/chart.reducer.ts
+++ b/packages/chart/src/store/chart.reducer.ts
@@ -28,6 +28,7 @@ type ChartState = {
 export const getInitialState = (configObj: ChartConfig): ChartState => {
   return {
     isLoading: true,
+
     config: defaults,
     stateData: _.cloneDeep(configObj?.data) || [],
     colorScale: null,
@@ -41,11 +42,7 @@ export const getInitialState = (configObj: ChartConfig): ChartState => {
     coveLoadedEventRan: false,
     isDraggingAnnotation: false,
     imageId: `cove-${Math.random().toString(16).slice(-4)}`,
-    brushConfig: {
-      data: [],
-      isActive: false,
-      isBrushing: false
-    }
+    brushData: []
   }
 }
 


packages/chart/src/CdcChartComponent.tsx
```diff
@@ -1100,7 +1100,13 @@ const CdcChart: React.FC<CdcChartProps> = ({
     setSharedFilterValue,
     svgRef,
     tableData: filteredData || excludedData,
-    transformedData: clean(filteredData || excludedData),
+    transformedData: clean(
+      Array.isArray(state.brushData) && state.brushData.length
```
adamdoe, line 1104: @Atash3000 won't this always be true? Do you have a way of checking if the brush chart is toggled on maybe?

packages/chart/src/components/LineChart/index.tsx
```diff
@@ -38,16 +38,12 @@ const LineChart = (props: LineChartProps) => {
     } = props
 
   // prettier-ignore
-  const { colorScale, config, formatNumber, handleLineType, parseDate, seriesHighlight, tableData, transformedData, updateConfig, brushConfig,clean  } = useContext<ChartContext>(ConfigContext)
-  const { yScaleRight } = useRightAxis({ config, yMax, data: transformedData, updateConfig })
+  const { colorScale, config, formatNumber, handleLineType, parseDate, seriesHighlight, tableData, transformedData:data, updateConfig,   } = useContext<ChartContext>(ConfigContext)
+  const { yScaleRight } = useRightAxis({ config, yMax, data: data, updateConfig })
```
adamdoe, line 42: You can just used `data` here over `data: data`

packages/chart/src/components/Brush/ParentBrush.tsx
adamdoe, line 1: This looks like the file for the actual brush controller that appears at the bottom of the screen. Maybe we should call this `BrushController.tsx` or if the controller can be used by multiple components `withBrushController.tsx`

packages/chart/src/components/LinearChart.tsx
```diff
@@ -402,7 +398,7 @@ const LinearChart = forwardRef<SVGAElement, LinearChartProps>(({ parentHeight, p
 
     // Heights to add
 
-    const brushHeight = brush?.active ? brush?.height + brush?.height : 0
+    const brushHeight = config.brush?.active ? 50 : 0
```
adamdoe, line 401: Assign 50 to a variable so we know why it was used in the future.

packages/chart/src/components/LineChart/index.tsx
```diff
@@ -38,16 +38,12 @@ const LineChart = (props: LineChartProps) => {
     } = props
 
   // prettier-ignore
-  const { colorScale, config, formatNumber, handleLineType, parseDate, seriesHighlight, tableData, transformedData, updateConfig, brushConfig,clean  } = useContext<ChartContext>(ConfigContext)
-  const { yScaleRight } = useRightAxis({ config, yMax, data: transformedData, updateConfig })
+  const { colorScale, config, formatNumber, handleLineType, parseDate, seriesHighlight, tableData, transformedData:data, updateConfig,   } = useContext<ChartContext>(ConfigContext)
```
adamdoe, line 41: Since you're changing from data: transformedData to transformedData: data, verify nothing is breaking in LineCharts.

packages/chart/src/components/LineChart/index.tsx
Atash3000: Fixed.

packages/chart/src/components/LinearChart.tsx
Atash3000: Fixed

packages/chart/src/components/Brush/ParentBrush.tsx
Atash3000: Fixed

PR #2221, opened 2025-07-07
[DEV-11002] Multiple Map Legend Color Fixes

adamdoe:
## Summary
* [DEV-11002]: On maps with filters color palettes generated wrong colors after filtering.
* [DEV-11094]: In certain situations bins had floating point artifacts which caused colors to not be colored in. ie. 35.800000004 instead of 35.8
* [DEV-11093]: When legend > separate zero was checked and use new quantile legend was being used, bin numbers were repeated, ie. 0, 1-10, 10-12 instead of 0, 1-10, 11-12
* [DEV-11095]: When special class was toggled to the end of the legend area the colors were displaying wrong, ie white instead of the first color in the color palette.
* [DEV-11096]: When special classes were set to show last, toggling the special class toggled the first color on the map instead of the special class.

### DIFF
--- a/packages/map/src/components/Legend/components/Legend.tsx
+++ b/packages/map/src/components/Legend/components/Legend.tsx
@@ -130,9 +130,11 @@ const Legend = forwardRef<HTMLDivElement, LegendProps>((props, ref) => {
 
         return runtimeLegend.items.findIndex(runtimeItem => {
           if (item.special && runtimeItem.special) {
-            // For special classes, match by value or label
-            return runtimeItem.value === (runtimeItem.label || runtimeItem.value) ||
-              runtimeItem.label === (item.label?.props?.children || item.label)
+            // For special classes, match by label (since formatted item label comes from runtime item)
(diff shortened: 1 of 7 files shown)


packages/map/src/components/Legend/components/Legend.tsx
```diff
@@ -313,41 +334,41 @@ const Legend = forwardRef<HTMLDivElement, LegendProps>((props, ref) => {
...
-                            </Group>
-                          </svg>
-                          <p>{label}</p>
-                        </div>
-                      )
-                  )}
-                </div>
-              </>
-            )}
+                <>
```
adamdoe, line 337: Formatting

Atash3000:
When legend > separate zero was checked and use new quantile legend was being used, bin numbers were repeated, ie. 0, 1-10, 10-12 instead of 0, 1-10, 11-12 and that it is still repeating ? 
Screenshot 2025-07-14 at 11 03 14

Atash3000 (approved):

PR #2224, opened 2025-07-09
FIX : [DEV-8890] - Fix Line chart tooltip hovers for single series

Atash3000:
## Testing Steps
Open Line charts
Add Single series tooltips ON from visual panel
Hover on line points. they suppose to show tooltip for each point.


adamdoe:
@Atash3000 did you regression test bar charts and the stacked bar charts for this PR?

PR #2239, opened 2025-07-21
Markup Inc. > No Data Message Enhancement

adamdoe:
## Summary
Adds a checkbox for stating "No Data Available" when the markup include has nothing to show.

## Testing Steps
Open the attached example `no-data-markup.json` and change the filter to something that doesn't match.

### DIFF
--- a/packages/dashboard/examples/no-data-markup.json
+++ b/packages/dashboard/examples/no-data-markup.json
@@ -1,0 +1,206 @@
+{
+  "dashboard": {
+    "theme": "theme-blue",
+    "sharedFilters": [
+      {
+        "key": "New Dashboard Filter 1",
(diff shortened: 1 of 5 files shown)


Atash3000:
@adamdoe 
Screenshot 2025-07-30 at 14 23 21
if bars has no data we should not see markup include ? it still shows 2021 . the checkbox is On.

adamdoe:
@Atash3000 this is for markup includes. 

On a dashboard:
- Add a filtered dropdown
- Add a markup include to change the markup include
- On the markup include you should be able to add a condition that doesn't meet the filters criteria. At that point "No Data" will be shown in the markup include (see the example file)

Atash3000 (approved):
Looks good!

PR #2274, opened 2025-08-15
fix: migration issue

Mgetz10:
## Summary
filterControlsStatePicked -> filterControlsStatesPicked

### Storybook Links
story/components-pages-dashboard--single-state-dashboard-with-filters


Mgetz10:
@adamdoe I noticed that too but I checked the last tag and i believe its a pre-existing issue

PR #2297, opened 2025-09-05
fix: state picked filtering logic

adamdoe:
This pull request makes a couple of targeted updates to the dashboard and map helper logic, primarily focusing on changing the dashboard configuration and correcting how states are picked in the map component.

Configuration update:

* The dashboard now loads its configuration from `/examples/private/north-dakota.json` instead of `/examples/private/oral-health.json` in `index.html`, which changes the dataset or region displayed on the dashboard.

Map helper logic correction:

* In `map.ts`, the `getFilterControllingStatesPicked` function now correctly uses `state.data` instead of `runtimeData` when mapping state filter selections, ensuring the filter logic references the intended data structure.

### DIFF
--- a/packages/dashboard/index.html
+++ b/packages/dashboard/index.html
@@ -22,7 +22,7 @@
 </head>
 
 <body>
-  <div class="react-container" data-config="/examples/private/oral-health.json"></div>
+  <div class="react-container" data-config="/examples/private/north-dakota.json"></div>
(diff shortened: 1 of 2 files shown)


adamdoe:
@Mgetz10 an issue bubbled up on dashboards with filters where the map wouldn't update because the runtimeData didn't include the state name in this area. Can you check that the multi single state maps are still working as intended?

PR #2333, opened 2025-09-24
Fix [DEV-11293] Standalone builds

jayb:
## Summary

Some more updates to make sure all packages can be built in isolation.


packages/chart/src/types/ChartConfig.ts
```diff
@@ -27,8 +27,8 @@ import { VizFilter } from '@cdc/core/types/VizFilter'
 import { type Annotation } from '@cdc/core/types/Annotation'
 import { Version } from '@cdc/core/types/Version'
 import Footnotes from '@cdc/core/types/Footnotes'
+import { ViewportSize } from '@cdc/core/types/ViewportSize'
```
adamdoe, line 30: Remove if unused

adamdoe (approved):

PR #2336, opened 2025-09-24
Fix/dev 11507

adamdoe:
On dashboards users couldn't choose:

Vertical > Multiple Rows > and select columns


packages/dashboard/src/components/DataDesignerModal.tsx
```diff
@@ -1,210 +1,220 @@
...
+  const configureData = useMemo(() => {
+    let baseConfig
+    if (isUpdatingVisualization) {
+      baseConfig = config.visualizations[vizKey]
+    } else {
+      baseConfig = config.rows[rowIndex]
+    }
+
+    // Directly attach data from datasets
+    const data = baseConfig?.dataKey ? config.datasets[baseConfig.dataKey]?.data : undefined
```
adamdoe, line 45: This is the change, it looks like prettier formatted the file when pushing.

PR #2733, opened 2026-04-16
Add config docs and maintenance workflows

jayb:
## Summary

Add config docs and maintenance workflows.


adamdoe (approved):
These additions look great. I didn't run the workflow or smoke tests, but I see this as a much-needed enhancement. Thanks!

PR #2760, opened 2026-04-23
DEV-12107 State Zoom failsafe when unexpected state column name used, and add "jurisdiction" as column name

khartco:
## Summary
Had a case where a dataset was using an unexpected column name for the state names that caused the data table to filter to nothing when zoomed in. This change makes it so the data table remains unchanged in such cases so it at least shows something. Also added the new new column name used, "jurisdiction".

Some of this might be improved more in the future, but this is a more minimal fix.

## Testing Steps
Load a county-level us map states under an unexpected column name.

### Storybook Links
n/a

### Screenshots
n/a

### DIFF
--- a/packages/map/src/helpers/dataTableHelpers.ts
+++ b/packages/map/src/helpers/dataTableHelpers.ts
@@ -23,8 +23,8 @@ export const filterCountyTableRuntimeDataByStateCode = (runtimeData: any, stateC
   const runtimeKeys = Object.keys(runtimeData)
   if (runtimeKeys.length === 0) return runtimeData
 
-  const stateName = supportedStateFipsCodes[stateCode]
-  const stateAbbreviation = stateFipsToAbbreviation[stateCode]
+  const stateName = supportedStateFipsCodes?.[stateCode]
+  const stateAbbreviation = stateFipsToAbbreviation?.[stateCode]
   const normalizedSelectedStateCode = String(stateCode).replace(/^0+/, '')
   const paddedSelectedStateCode = normalizedSelectedStateCode.padStart(2, '0')
   const stateColumnNames = Object.values(config?.columns || {})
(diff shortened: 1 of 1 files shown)


packages/map/src/helpers/dataTableHelpers.ts
```diff
@@ -13,18 +13,46 @@ export const shouldShowDataTable = (config: any, table: any, general: any, loadi
...
  */
 export const filterCountyTableRuntimeDataByStateCode = (runtimeData: any, stateCode: string, config?: any) => {
   if (!runtimeData || runtimeData.init || !stateCode) return runtimeData
 
-  const filtered = {}
-  const stateName = supportedStateCodes[stateCode]
+  const runtimeKeys = Object.keys(runtimeData)
+  if (runtimeKeys.length === 0) return runtimeData
+
+  const stateName = supportedStateFipsCodes[stateCode]
```
adamdoe: If there's a chance of stateCode not existing in supportedStateFipsCodes use `.?[stateCode]`

packages/map/src/helpers/dataTableHelpers.ts
```diff
@@ -13,18 +13,46 @@ export const shouldShowDataTable = (config: any, table: any, general: any, loadi
...
   const stateAbbreviation = stateFipsToAbbreviation[stateCode]
   const normalizedSelectedStateCode = String(stateCode).replace(/^0+/, '')
   const paddedSelectedStateCode = normalizedSelectedStateCode.padStart(2, '0')
   const stateColumnNames = Object.values(config?.columns || {})
     .map((column: any) => column?.name)
-    .filter((name: string) => !!name && /(state|territory|fips)/i.test(name))
+    .filter((name: string) => !!name && /(state|territory|fips|jurisdiction)/i.test(name))
+
+  // Also check common state field names directly in the data rows
+  const commonStateFieldNames = [
```
adamdoe, line 35: Instead of guessing the column names, can you use what the user has already defined?

```.js
const geoColumnName = config?.columns?.geo?.name`
const allStateColumns = [...new Set([...stateColumnNames, geoColumnName])]
```

adamdoe (changes requested):

adamdoe (approved):

PR #2789, opened 2026-05-05
Fix: Chart footnotes

jayb:
## Summary

Fixes a bug where chart footnotes cannot be added or updated in the editor. Seems like it's somehow been around since daa9dade7?

This PR also introduces [s]css formatting with Prettier on commit, hopefully we get some more consistency there.

## Testing Steps

Open a chart and add a footnote. It should display on the chart.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -7,5 +7,6 @@
   "bracketSameLine": false,
   "arrowParens": "avoid",
   "trailingComma": "none",
-  "printWidth": 120
+  "printWidth": 120,
+  "editorconfig": true
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -7,6 +7,5 @@
   "bracketSameLine": false,
   "arrowParens": "avoid",
   "trailingComma": "none",
-  "printWidth": 120,
-  "editorconfig": true
+  "printWidth": 120
 }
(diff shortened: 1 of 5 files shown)


.prettierrc
```diff
@@ -7,6 +7,5 @@
   "bracketSameLine": false,
   "arrowParens": "avoid",
   "trailingComma": "none",
-  "printWidth": 120,
-  "editorconfig": true
```
adamdoe, line 11: It looks like this can be dropped starting in v3 of prettier, however, it looks like we're still on version 2. Verify options remain working.

package.json
```diff
@@ -104,7 +104,11 @@
       "npx prettier --write"
     ],
     "packages/**/*.css": [
-      "stylelint --fix"
+      "stylelint --fix",
```
adamdoe, line 107: good: applies formatting for files we had missing.

packages/chart/src/CdcChartComponent.tsx
```diff
@@ -1431,7 +1431,7 @@ const CdcChart: React.FC<CdcChartProps> = ({
             }
             footer={
               <FootnotesStandAlone
-                config={configObj.footnotes}
+                config={config.footnotes}
```
adamdoe, line 1434: good: config is the property managed by useReducer here

adamdoe (approved):

PR #2822, opened 2026-05-18
Fix [DEV-12960] Limit width of filters when note is present

jayb:
## Summary

Limits the width of filters when a note is present.

## Testing Steps

Open [filter-section-styling.json](https://github.com/user-attachments/files/27961459/filter-section-styling.json) as a dashboard config. The note should be wider than the filter.


adamdoe (approved):
Looks good!

PR #2827, opened 2026-05-19
DEV-9267 Add map filter clear selection control

khartco:
## Summary
When the US map is used for filtering a dashboard, user is now able to select and deselect the same state to clear a selection. In addition, utilizes the zoom controls "clear selection" button as an optional button thats on by default that also clears the selection.

## Testing Steps
Dashboard with us map as filter (a config is available on the DEV-9267 ticket)


adamdoe (changes requested):
fix: I think scoping the filter-clearing to the zoom component feels wrong, but happy to discuss further. Can this functionality be moved outside of zooming? If a specific map type doesn't support zooming but does support filtering, there might be a mismatch in functionality for the user.

PR #2856, opened 2026-05-27
Fix for date-axis in scatter plots

khartco:
## Summary
Choosing date scaling type "Date" (any) for scatter plots is resulting in the plots being drawn improperly. "Categorical" works fine though. This should fix it to allow both types. Determined this should go in the next release and not squeeze into the one about to go out.

## Testing Steps
Scatter plot > Editor > Date/Category Axis > "Date" or "Categorical"


adamdoe (approved):
Tested multiple files and legend types, looks good!

PR #2883, opened 2026-06-12
Dev 13164

Viseghoh:
## Summary
Host a feedback changes for heatmap.

- Adjust font sizes for Axis Labels and Tick Labels
- Additional Tick Rotation
- Value Ranges added to Gradient
- Prefix and Suffix fixes for Data Series


packages/chart/src/components/HeatMap/components/HeatMap.tsx
```diff
@@ -64,7 +64,7 @@ type HeatMapLayout = {
 const AXIS_TOP_WITH_TICKS = 36
 const AXIS_TOP_WITHOUT_TICKS = 24
 const AXIS_TOP_LABEL_SPACE = 28
-const AXIS_TICK_FONT_SIZE = 12
+const AXIS_TICK_FONT_SIZE = 16
```
adamdoe, line 67: @Viseghoh some of these settings are saved as css variables. It might be a good idea to move towards using those to avoid differences between chart types.

packages/chart/src/components/EditorPanel/helpers/getVisualizationTypeConfigUpdate.ts
```diff
@@ -22,7 +22,11 @@ export const getVisualizationTypeConfigUpdate = (
     return {
       ...config,
       visualizationType,
-      series: [],
+      series: (config.series || []).map(series => ({
+        ...series,
+        type: 'HeatMap',
```
adamdoe, line 27: This looks like it will set every series type to heatmap. If this isn't used outside of heatmaps thats fine but the name of the file makes me feel like this should be more generic.

adamdoe (approved):

PR #2956, opened 2026-07-16
DEV-13355 Update table screen reader so it only reads once for headers

khartco:
## Summary
Update the aria label for table headers so it's only read when the table header is highlighted, and not when the table cells below are highlighted. This ended up being a little tricky as by default the entire content of the header is read, so this updates depending on whats highlighted or not.

Rebased for test from dev

## Testing Steps
Checking with screenreader on any datatable such as the default chart should work as expected.


Viseghoh (approved):
I tried it out and it seems to work fine! I only it heard it describe the header once in storybook testing.

PR #2996, opened 2026-08-06
enhancement: auto adjust right axis

adamdoe:
This pull request improves the handling of right Y-axis sizing in linear charts by dynamically measuring and enforcing a minimum gutter width to prevent visual overlap or clipping of axis content. The main changes include introducing a new helper to calculate the right Y-axis width, updating the chart layout calculation to use this measured width, and enhancing documentation to clarify runtime behavior.

**Right Y-axis sizing improvements:**

* Added a new helper function `calculateRightYAxisWidth` to measure the required gutter for the right Y-axis based on tick labels, side titles, and configured minimums, ensuring right-axis content does not overlap legends or spill outside the chart area.
* Updated `LinearChart.tsx` to use the measured right Y-axis width in layout calculations, replacing the previous approach that relied solely on the configured `rightAxisSize`. [[1]](diffhunk://#diff-037ed66da5b35c536d9df7411ecb790cb4566cd0e50a0c9f33d9db5574c483bfR40) [[2]](diffhunk://#diff-037ed66da5b35c536d9df7411ecb790cb4566cd0e50a0c9f33d9db5574c483bfR292-R304)

**Testing:**

* Added a comprehensive test suite for `calculateRightYAxisWidth` to verify correct gutter calculation under various right-axis configurations and edge cases.

**Documentation:**

* Updated `CONFIG.md` to clarify that when right-axis content is visible, the runtime uses the authored `rightAxisSize` as a minimum gutter but enforces a measured minimum to prevent content overlap.


adamdoe:
@jayb @Mgetz10 - can you take a look here if time allows? I remember work being completed for the left axis auto widths. This seems to fix the right now. Thanks!

jayb (commented):
@adamdoe This works really well. A couple minor thoughts:

When we did the left axis auto-sizing, we made it so the value in the config no longer has an impact, and the control was removed from the editor. Here it looks like we still partially support `yAxis.rightAxisSize` as a minimum value. Is there a specific reason we need to keep this partial/minimum support, or could we remove the control from the editor and ignore the `yAxis.rightAxisSize` value in configs? If we need to keep it, maybe the editor label should be changed to "Minimum Size" so it's clearer.

The spacing could be cleaned up a little bit:

* The right axis content should always be tight to the right side of the chart. When there's no axis label, there's a bit of extra space after the tick labels

* The distance between the right side of tick labels and a side-positioned axis label is visually a bit smaller than the spacing on the left axis.

Screenshots and config:

Image

Image

[measlesvax.json](https://github.com/user-attachments/files/30903797/measlesvax.json)

adamdoe:
@jayb - can you run another cycle here?

jayb (approved):
Looks good!

PR #3004, opened 2026-08-14
fix: prefix suffix and round

adamdoe:


Viseghoh (approved):
Tested and confirmed. Good work @adamdoe !

PR #3011, opened 2026-08-21
Add Custom Label Condition

Viseghoh:
## Summary
Tweaked the logic so that a custom column label will have the highest priority when the viz is a HeatMap or the column belongs to a user defined series.

- Updated getSeriesName to return the canonical series name without applying Column Label overrides.
- Added getDataTableColumnLabel to handle Data Table-specific labels.
- Standardized legend and series tooltip labels to use the same series-name fallback logic.
- Preserved Column Labels for Data Table headers and additional non-series tooltip fields.
- Preserved HeatMap’s existing Column Label behavior.
- Added regression tests and updated CONFIG.md to document t


packages/core/components/DataTable/helpers/getSeriesName.ts
```diff
@@ -15,17 +22,18 @@ const getLabel = (name: string, config: TableConfig) => {
 
 export const getSeriesName = (column: string, config: TableConfig) => {
```
adamdoe: It seems like `getSeriesName` should return the true series name. If something needs to override that, can we update where it's called instead?  Maybe, `getUserDefinedSeries` is better. This just feels like an opportunity where we might want to keep both functions, one for the actual series name, and one for the user defined series name.

adamdoe (changes requested):

packages/core/components/DataTable/helpers/getSeriesName.ts
Viseghoh: @adamdoe ya that's a good call. I'm fixing it a bit now and will re-push here shortly.

Viseghoh: @adamdoe getSreisName should only get the canonical series name now. I added a method specifically for the Data Table to column labels. So now the Legend and Series Tooltips should look directly at series name for labeling. Data Table should look at the column label settings first and then fall back to series name. I figured having 2 clearly named helpers was beter than passing flags and into getSeriesName so its easier to test.

khartco (approved):

adamdoe (approved):

PR #3022, opened 2026-08-27
Fix/auto align question

adamdoe:
@jayb - this will probably be a good one for you to review. 

A requestor reported that stacked, isolated bars were visible even at very low values (e.g., .005) when "align axis" was disabled, thanks to a 3px minimum height flag. However, when "align to axis" was enabled, that 3px minimum height no longer applied, and the bars didn't show at all.


jayb (approved):
@adamdoe Looks good -- I tested it and saw the new behavior, and the changes makes sense to me. Thanks for fixing!

PR #3023, opened 2026-08-31
Dev 10599

Viseghoh:
## Summary

Percentage-decorated values such as `12.5%` were treated as nonnumeric, preventing charts and map legends from using them even though the data table displayed them correctly.

### What changed

- Bar charts remove a trailing `%` for calculations while preserving it for tooltip display.
- Suppression values remain intact during data cleaning.
- Map legends infer `%` only when all active numeric values are percentages.
- Percentage labels work across calculated ranges, manual ranges, and separate-zero legends.
- Existing configured suffixes and mixed percentage/regular-number datasets remain unchanged.

### Behavior impact

This is presentational only. Values, scales, sorting, map bins, and calculations behave as before—`12.5%` is calculated as `12.5`, not `0.125`.

### Coverage

Tests cover vertical, horizontal, stacked, dynamic, brushed, duplicate-category, and suppressed bar data, plus filtered, mixed-unit, special-class, and zero-value map legends.

Most of the PR size comes from regression tests and Storybook coverage rather than production logic.


packages/chart/src/CdcChartComponent.tsx
```diff
@@ -1351,16 +1385,16 @@ const CdcChart: React.FC<CdcChartProps> = ({
...
   const configYAxisDomainData = (config as ChartConfig).yAxisDomainData
   const yAxisDomainData = useMemo(() => {
     if (Array.isArray(configYAxisDomainData) && configYAxisDomainData.length > 0) {
-      return clean(getExcludedData(config, configYAxisDomainData))
+      return cleanChartData(getExcludedData(config, configYAxisDomainData))
     }
 
-    return clean(excludedData)
+    return cleanChartData(excludedData)
   }, [config, configYAxisDomainData, excludedData])
```
Viseghoh, line 1398: Reviewed, this was pre-existing befoer this ticket. We only renamed cleanData to cleanChartData.

Viseghoh:
@adamdoe @khartco if you have bandwith to take a peek this would be appreciated. It looks like a lot but it was mostly storybook testing.

khartco (approved):

adamdoe (approved):

PR #3060, opened 2026-09-17
fix: show all territories on geocode maps

adamdoe:


Viseghoh (approved):
@adamdoe tested it out with the geocode example map, looks good.

